PR #1, opened 2024-05-21
doc: add design doc for the ceph-csi-operator

Madhu-1:
Adding intial design doc for the ceph-csi-operator

### DIFF
--- a/docs/design/operator.md
+++ b/docs/design/operator.md
@@ -1,0 +1,388 @@
+# Ceph CSI Operator Design Document
+
+## Introduction
+
+This document outlines the design of the Ceph-CSI Operator, which aims to
+provide Kubernetes native management interfaces for Ceph-CSI drivers (CephFS,
+RBD, and NFS) for Kubernetes based environments. The operator automates the
+deployment, configuration, and management of these drivers using new Kubernetes
+APIs defined as a set of Custom Resource Definitions (CRDs).
+
+## System Overview
+
+The Ceph-CSI Operator is a Kubernetes operator that offers a Kubernetes native
+management interfaces for Ceph CSI drivers. It provides a set of new Kubernetes
+APIs (CRDs) for governing deployment and management of Ceph-CSI drivers, as
+well as connection details to Ceph clusters. The operator ensures that the
+desired state specified in the CRDs is maintained in the cluster.
+
+## Assumptions and Dependencies
+
+Kubernetes cluster is available and operational. Ceph cluster is set up and
+accessible from the Kubernetes cluster. Users are familiar with Kubernetes and
+Ceph concepts. CRD and operator versions are compatible with the Kubernetes
+version in use.
+
+## Goals and Guidelines
+
+- Automate the deployment and lifecycle for Ceph-CSI drivers.
+- Provide a consistent and flexible configuration mechanism using CRDs.
+- Enable namespace-scoped configurations for different Ceph-CSI drivers.
+- Ensure high availability and scalability of the operator.
+- Support multiple deployment modes for the operator:
+  - Own-namespace: the operator manages a single Ceph-CSI driver in the current
+    namespace where operator is deployed.
+  - Multi-namespace: the operator manages Ceph-CSI drivers in user-specified
+    namespaces.
+  - All-namespaces: the operator manages Ceph-CSI drivers in all namespaces.
+
+## Development Methods
+
+Use kubebuilder framework which is highly used in operator development,
+following Kubernetes best practices for controllers and operators
+
+## Architecture
+
+```mermaid
+graph TD
+    A[CRD Changes] --> B[Operator]
+    B -->  D[Configure CephFS CSI Driver]
+    B -->  E[Configure RBD CSI Driver]
+    B -->  F[Configure NFS CSI Driver]
+
+```
+
+In this diagram:
+
+- **CRD Changes**: Represents changes made to Custom Resource Definitions,
+  which trigger actions in the operator.
+- **Operator**: Listens for changes in CRDs and initiates the installation and
+  configuration of CSI drivers.
+- **Configure CephFS, NFS, RBD**: Actions performed by the operator to install
+  and configure the respective CSI drivers based on the CRD changes.
+
+## CRDs for ceph-csi-operator
+
+### CephCSIOperatorConfig CRD
+
+Manages operator-level configurations and offers a place to overwrite settings
+for CSI drivers. This CRD is a namespace-scoped CRD and a single CR named
+ceph-csi-operator-config can be created by an admin inside the operator's
+namespace.
+
+The configurations are catagorized into 2 different types
+
+- Operator configurations
+- Default configuration overwrites applied to all Ceph CSI drivers
+
+```yaml
+---
+kind: CephCSIOperatorConfig
+apiVersion: csi.ceph.io/v1alpha1
+metadata:
+  name: ceph-csi-operator-config
+  namespace: <operator-namespace>
+spec:
+  logLevel: 1
+  driverSpecDefaults:
+    logging:
+      logLevel: 5
+      maxfiles: 5
+      maxLogSize: 10M
+    clusterName: 5c63ad7e-74fe-4724-a511-4ccdc560da56
+    enableMetadata: true
+    grpcTimeout: 100
+    snapshotPolicy: auto-detect
+    generateOMapInfo: true
+    fsGroupPolicy: File
+    encryption:
+      configMapRef:
+        name: encryption-config-map-name
+    plugin:
+      priorityClassName: system-node-critical
+      updateStrategy:
+        type: RollingUpdate
+        rollingUpdate:
+          maxUnavailable: 1
+      labels:
+        app: csi
+      annotations:
+        k8s.v1.cni.cncf.io/networks: macvlan-conf-1
+      affinity:
+        nodeAffinity:
+          requiredDuringSchedulingIgnoredDuringExecution:
+            nodeSelectorTerms:
+            - matchExpressions:
+              - key: storage
+                operator: In
+                values:
+                - node
+      tolerations:
+      - key: storage
+        operator: Exists
+      resources:
+        registrar:
+          limits:
+            cpu: '200'
+            memory: '500'
+          requests:
+            cpu: '100'
+            memory: '250'
+        liveness:
+          limits:
+            cpu: '200'
+            memory: '500'
+          requests:
+            cpu: '100'
+            memory: '250'
+        plugin:
+          limits:
+            cpu: '200'
+            memory: '500'
+          requests:
+            cpu: '100'
+            memory: '250'
+      volumes:
+      - name: host-run
+        volumes:
+          hostPath:
+            path: "/run"
+            type: Directory
+        volumeMounts:
+          name: ''
+          readOnly: true
+          mountPath: "/run"
+          mountPropagation: Bidirectional
+      kubeletDirPath: "/var/lib/kubelet"
+      imagePullPolicy: IfNotPresent
+    provisioner:
+      priorityClassName: system-cluster-critical
+      labels:
+        app: provisioner
+      annotations:
+        k8s.v1.cni.cncf.io/networks: macvlan-conf-1
+      provisionerReplicas: 2
+      affinity:
+        nodeAffinity:
+          requiredDuringSchedulingIgnoredDuringExecution:
+            nodeSelectorTerms:
+            - matchExpressions:
+              - key: storage
+                operator: In
+                values:
+                - node
+      tolerations:
+      - key: storage
+        operator: Exists
+      resources:
+        attacher:
+          limits:
+            cpu: '200'
+            memory: '500'
+          requests:
+            cpu: '100'
+            memory: '250'
+        snapshotter:
+          limits:
+            cpu: '200'
+            memory: '500'
+          requests:
+            cpu: '100'
+            memory: '250'
+        resizer:
+          limits:
+            cpu: '200'
+            memory: '500'
+          requests:
+            cpu: '100'
+            memory: '250'
+        provisioner:
+          limits:
+            cpu: '200'
+            memory: '500'
+          requests:
+            cpu: '100'
+            memory: '250'
+        omapGenerator:
+          limits:
+            cpu: '200'
+            memory: '500'
+          requests:
+            cpu: '100'
+            memory: '250'
+        liveness:
+          limits:
+            cpu: '200'
+            memory: '500'
+          requests:
+            cpu: '100'
+            memory: '250'
+        plugin:
+          limits:
+            cpu: '200'
+            memory: '500'
+          requests:
+            cpu: '100'
+            memory: '250'
+    liveness:
+      metricsPort: 8000
+    leaderElection:
+      leaseDuration: 100
+      renewDeadline: 100
+      retryPeriod: 10
+    deployCSIAddons: true
+    cephFsKernelClient: true
+    cephFsKernelMountOptions: ms_mode=secure
+status:
+  phase: Succeeded
+  reason: operator config successfully created
+```
+
+### CephCSIDriver CRD
+
+Manages the installation, lifecycle management, and configuration for CephFS,
+RBD, and NFS CSI drivers within namespaces.
+
+- Allows customization of driver settings on a per-driver basis.
+- Best practice dictates only a single instance of RBD/NFS/CephFS drivers
+  should be created in a single namespace. If there is a need to deploy a
+  second instance of a driver of a similar type, it should be deployed in a
+  different namespace.
+- Each CSI driver must have a unique name across all namespaces.
+
+```yaml
+---
+kind: CephCSIDriver
+apiVersion: csi.ceph.io/v1alpha1
+metadata:
+  name: "<prefix>.<driver_type>.csi.ceph.com"
+  namespace: <operator-namespace>
+spec:
+  fsGroupPolicy: File
+  encryption:
+    configMapRef:
+      name: encryption-config-map-name
+  plugin:
+    priorityClassName: system-node-critical
+    updateStrategy:
+      type: RollingUpdate
+      rollingUpdate:
+        maxUnavailable: 1
+    labels:
+      app: cephfs-plugin
+    annotations:
+      k8s.v1.cni.cncf.io/networks: macvlan-conf-1
+  provisioner:
+    labels:
+      app: ceph-fs-provisioner
+    annotations:
+      k8s.v1.cni.cncf.io/networks: macvlan-conf-1
+    provisionerReplicas: 2
+    leaderElection:
+      leaseDuration: 100
+      renewDeadline: 100
+      retryPeriod: 10
+    attachRequired: true
+  liveness:
+    metricsPort: 8000
+  deployCSIAddons: false
+  kernelClient: true
+  kernelMountOptions: ms_mode=secure
+status:
+  phase: Failed
+  reason: csi driver with same name already exists in the cluster
+```
+
+### CephCSICephCluster CRD
+
+Stores connection and configuration details for a single Ceph cluster and
+provide the information to be used by multiple CSI drivers.
+
+```yaml
+---
+kind: CephCSICephCluster
+apiVersion: csi.ceph.io/v1alpha1
+metadata:
+  name: ceph-cluster-1
+  namespace: <operator-namespace>
+spec:
+  monitors:
+  - 10.98.44.171:6789
+  - 10.98.44.172:6789
+  - 10.98.44.173:6789
+  ReadAffinity:
+    crushLocationLabels:
+    - kubernetes.io/hostname
+    - topology.kubernetes.io/region
+    - topology.kubernetes.io/zone
+  rbdMirrorDaemonCount: 2
+  cephConfig:
+    global:
+      auth_cluster_required: none
+      auth_service_required: none
+      auth_client_required: none
+      rbd_validate_pool: false
+status: {}
+```
+
+### CephCSIConfig CRD
+
+Contains details about CephFS, RBD, and NFS configuration to be used when
+communicating with Ceph. Include a reference to a CephCSICephCluster holding
+the connection information for the target Ceph cluster.
+
+```yaml
+---
+kind: CephCSIConfig
+apiVersion: csi.ceph.io/v1alpha1
+metadata:
+  name: storage
+  namespace: <operator-namespace>
+spec:
+  cephClusterRef:
+    name: ceph-cluster-1
+  cephFS:
+    subvolumeGroup: csi
+    kernelMountOptions: readdir_max_bytes=1048576,norbytes
+    fuseMountOptions: debug
+  rbd:
+    radosNamespace: rados-test
+status:
+  phase: Succeeded
+  reason: successfully linked to CephClusterRef
+```
+
+### CephCSIConfigMapping
+
+The CephCSIConfigMapping CR contains a mapping between local and remote Ceph
+cluster configurations.It also provides a mapping between the
+CephCSIConfigurations. The information in this CR helps Ceph-CSI identify
+peered blockpools. This can help with the correct management and consumption of
+volumes in backup and DR scenarios
+
+```yaml
+---
+kind: CephCSIConfigMapping
+apiVersion: csi.ceph.io/v1
+metadata:
+  name: storage
+  namespace: <operator-namespace>
+spec:
+  blockPoolMapping:
+  - local:
+      cephCSIConfigName: remote1-cephCSICluster-name
+      poolID: 2
+    remote:
+      cephCSIConfigName: remote1-cephCSICluster-name
+      poolID: 2
+  - local:
+      cephCSIConfigName: remote1-cephCSICluster-name
+      poolID: 2
+    remote:
+      cephCSIConfigName: remote2-cephCSICluster-name
+      poolID: 3
+```
+
+By following this design document, the Ceph CSI Operator can be effectively
+implemented, providing automated and scalable management of Ceph CSI drivers
+within Kubernetes clusters.


docs/design/operator.md
```diff
@@ -0,0 +1,414 @@
...
+- **Operator**: Listens for changes in CRDs and initiates the installation and
+  configuration of CSI drivers.
+- **Configure CephFS, NFS, RBD**: Actions performed by the operator to install
+  and configure the respective CSI drivers based on the CRD changes.
+
+## CRD's for ceph-csi-operator
+
+### CephCSIOperatorConfig CRD
+
+Manages operator-level configurations and default settings for CSI drivers.
```
nb-ohad, line 73:
```suggestion
Manages operator-level configurations and offers a place to overwrite default settings for CSI drivers.
```

docs/design/operator.md
```diff
@@ -0,0 +1,414 @@
...
+Ensures consistent default configurations across all CSI drivers.  This CRD is
```
nb-ohad, line 74: I am not sure that the above line  is true 
```suggestion
This CRD is
```

docs/design/operator.md
```diff
@@ -0,0 +1,414 @@
...
+a namespace scoped CRD and a single CR instance should be created by the admin
```
nb-ohad, line 75:
```suggestion
a namespace-scoped CRD and a single CR named ceph-csi-operator-config can be created by the admin 
```

docs/design/operator.md
```diff
@@ -0,0 +1,414 @@
...
+in the namespace where operator is running. The configurations are catagorized
+into 4 different parts
+
+- The operator configurations
```
nb-ohad, line 79:
```suggestion
- Operator configurations
```

docs/design/operator.md
```diff
@@ -0,0 +1,414 @@
...
+- Common configurations across all csi driver
+- Provisioner configurations
+- The Plugin Configurations
```
nb-ohad, line 82: These 2 are part of the common configuration, I am not sure why we need to differentiate them. 
```suggestion
```

docs/design/operator.md
```diff
@@ -0,0 +1,414 @@
+# Ceph CSI Operator Design Document
+
+## Introduction
+
+This document outlines the design of the Ceph-CSI Operator, which aims to
+provide easy management of Ceph-CSI drivers (CephFS, RBD, and NFS) in
```
nb-ohad, line 6:
```suggestion
provide Kubernetes native management interfaces for Ceph-CSI drivers (CephFS, RBD, and NFS) for
```

docs/design/operator.md
```diff
@@ -0,0 +1,414 @@
...
+Kubernetes based environments. The operator automates the deployment,
+configuration, and management of these drivers using Custom Resource
```
nb-ohad, line 8:
```suggestion
configuration, and management of these drivers using new Kubernetes APIs defined as a set of Custom Resource
```

docs/design/operator.md
```diff
@@ -0,0 +1,414 @@
...
+Definitions (CRDs).
+
+## System Overview
+
+The Ceph-CSI Operator is a Kubernetes operator that simplifies the management
```
nb-ohad, line 13: "Simplify" is an opinion, not a fact
```suggestion
The Ceph-CSI Operator is a Kubernetes operator that offers a Kubernetes native management interfaces for
```

docs/design/operator.md
```diff
@@ -0,0 +1,414 @@
...
+of Ceph CSI drivers. It provides a set of CRDs to manage the configurations of
```
nb-ohad, line 14:
```suggestion
of Ceph CSI drivers. It provides a set of new Kubernetes APIs (CRDs) for governing deployment and management of Ceph-CSI's drivers
```

docs/design/operator.md
```diff
@@ -0,0 +1,414 @@
...
+CephFS, RBD, and NFS drivers, as well as connection details to Ceph clusters.
```
nb-ohad, line 15:
```suggestion
, as well as connection details to Ceph clusters.
```

docs/design/operator.md
```diff
@@ -0,0 +1,414 @@
...
+The operator ensures that the desired state specified in the CRDs is maintained
+in the cluster.
```
nb-ohad, line 17: This is true for every operator, I am not sure it added anything. Maybe we want to remove this

docs/design/operator.md
```diff
@@ -0,0 +1,414 @@
...
+## Goals and Guidelines
+
+Automate the installation and management of Ceph-CSI drivers. Provide a
+consistent and flexible configuration mechanism using CRDs. Enable
+namespace-scoped configurations for different Ceph-CSI drivers. Ensure high
+availability and scalability of the operator. When running operator in a
+namespace provide an option to watch for the CR's in the  namespace where its
+running for ceph-CSI management. User will have an option to deploy multiple
+ceph-csi-operator that are configured to watch for the namespaces where they
+are deployed.
```
nb-ohad, line 42:
```suggestion

- Automate the deployment and lifecycle for Ceph-CSI drivers. Provide a
- Provide a consistent and flexible configuration mechanism using CRDs
- Enable namespace-scoped configurations for different Ceph-CSI drivers
- Ensure high availability and scalability of the operator
- Support multiple deployment methods for the operator (single-namespace, multi-namespace, all-namespaces)
-  Support multiple operator deployment in a single namespace when running in a single-namespace mode

File: docs/design/operator.md [hunk 13]
```diff
@@ -0,0 +1,414 @@
...
+
+## Development Methods
+
+Use operator framework which is higly used in operator developed and also by
+following Kubernetes best practices for operator development.
```
nb-ohad, line 47: I want to revisit this, I am now in favor of using `kubebuilder`

docs/design/operator.md
```diff
@@ -0,0 +1,414 @@
...
+            cpu: '100'
+            memory: '250'
+        plugin:
+          limits:
+            cpu: '200'
+            memory: '500'
+          requests:
+            cpu: '100'
+            memory: '250'
+      pluginVolumes:
```
nb-ohad, line 151: We are already under the plugin section
```suggestion
     volumes:
```

docs/design/operator.md
```diff
@@ -0,0 +1,414 @@
...
+      - name: host-run
+        Volumes:
```
nb-ohad, line 153:
```suggestion
        volumes:
```

docs/design/operator.md
```diff
@@ -0,0 +1,414 @@
...
+          hostPath:
+            path: "/run"
+            type: Directory
+        VolumeMounts:
```
nb-ohad, line 157:
```suggestion
        volumeMounts:
```

docs/design/operator.md
```diff
@@ -0,0 +1,414 @@
...
+    liveness:
+      metricsPort: 8000
+    leaderElection:
+      leaseDuration: 100
+      renewDeadline: 100
+      retryPeriod: 10
+    deployCSIAddons: true
+    cephfs:
+      useKernelClient: true
+      kernelMountOptions: ms_mode=secure
```
nb-ohad, line 242: We agreed to not use a dedicated section  
```suggestion
          cephFsKernelClient: true
          cephFsKernelMountOptions: ms_mode=secure
```

docs/design/operator.md
```diff
@@ -0,0 +1,414 @@
...
+status:
+  phase: Succeeded
+  message: operator config successfully created
```
nb-ohad, line 245: Isn't the common practice to use a `reason` filed instead of a `message` field?
```suggestion
  message: operator config successfully created
```

docs/design/operator.md
```diff
@@ -0,0 +1,414 @@
...
+```
+
+### CephCSIDriver CRD
+
+Manages the installation and configuration of CephFS, RBD, and NFS CSI drivers
+within namespaces. Allows customization of driver settings on a per-namespace
+basis. Only single instance of RBD/NFS/CephFS drivers can be created in a
+single namespace, If user wants to deploy new instance of RBD it need to be
+created in a new namespace. The operator need to ensure that no two CR'S are
+created in different namespace that are targetting same csi driver name.
```
nb-ohad, line 255:
```
Only single instance of RBD/NFS/CephFS drivers can be created in a
single namespace
```

I do not agree with this restriction, there is no need for it.


```suggestion
Manages the deployment and lifecycle of a single CephFS, RBD, or NFS CSI driver.
Allows customization of driver settings on a per-driver basis. 
```

docs/design/operator.md
```diff
@@ -0,0 +1,414 @@
...
+
+```yaml
+---
+kind: CephCSIDriver
+apiVersion: csi.ceph.io/v1alpha1
+metadata:
+  name: "<prefix>.cephfs.csi.ceph.com"
```
nb-ohad, line 262:
```suggestion
  name: "<driver_name>.<driver_type>.csi.ceph.com"
```

docs/design/operator.md
```diff
@@ -0,0 +1,414 @@
...
+        leaseDuration: 100
+        renewDeadline: 100
+        retryPeriod: 10
+      attachRequired: true
+    liveness:
+      metricsPort: 8000
+    deployCSIAddons: false
+  cephfs:
+    useKernelClient: true
+    kernelMountOptions: ms_mode=secure
```
nb-ohad, line 298: We agreed to not use a dedicated section  
```suggestion
          cephFsKernelClient: true
          cephFsKernelMountOptions: ms_mode=secure
```

docs/design/operator.md
```diff
@@ -0,0 +1,414 @@
...
+status:
+  phase: Failed
+  message: Failed to create cephfs csi driver
```
nb-ohad, line 301: Why do we need both `message` and `reason`?

docs/design/operator.md
```diff
@@ -0,0 +1,414 @@
...
+
+### CephCSIClusterRecovery
+
+The CephCSIClusterRecovery CR contains the local and the remote cephcluster
+configuration which will help cephcsi to identify the blockpool alias to
+consume the volume from when doing CSI operation.
+
+```yaml
+---
+kind: CephCSIClusterRecovery
```
nb-ohad, line 371: We should not tie this CR to the realm of Recovery. It might be useful for other use cases in the future.
```suggestion
kind: CephCSIClusterMapping
```

docs/design/operator.md
```diff
@@ -0,0 +1,414 @@
...
+apiVersion: csi.ceph.io/v1alpha1
+metadata:
+  name: storage
+  namespace: operator-namespace
+spec:
+  blockPoolMapping:
+  - local:
+      cephCSIClusterConfigRef:
+        name: remote1-cephCSICluster-name
```
nb-ohad, line 380: What is the reason to use a local object ref here? It is either way pointing to a config name, not a config CR.
For remote it makes even less sense to use local object ref as there is no local resource for this name.

Can we just use a clusterConfigName field instead?

docs/design/operator.md
```diff
@@ -0,0 +1,414 @@
...
+```
+
+### CephCSIClusterConfig CRD
+
+Contains details about CephFS subvolume group or RADOS namespace and a
+reference to CephCSICephConfig to link storage configurations to Ceph cluster.
+
+```yaml
+---
+kind: CephCSIClusterConfig
```
nb-ohad, line 348: Cluster is creating confusion here as Ceph itself has a cluster in a meaningful and different way and we already have a `CephCSICephCluster` CR in place. I believe we agreed on `CephCSIConfig` in our offline discussion
```suggestion
kind: CephCSIClusterConfig
```

docs/design/operator.md
```diff
@@ -0,0 +1,414 @@
...
+apiVersion: csi.ceph.io/v1alpha1
+metadata:
+  name: storage
+  namespace: operator-namespace
+spec:
+  cephCSICephConfigRef:
```
nb-ohad, line 354: We are already inside a ceph-csi context, I am not sure we need to repeat it again and again. 
```suggestion
  cephClusterRef:
```

docs/design/operator.md
```diff
@@ -0,0 +1,414 @@
...
+  - 10.98.44.172:6789
+  - 10.98.44.173:6789
+  ReadAffinity:
+    crushLocationLabels:
+    - kubernetes.io/hostname
+    - topology.kubernetes.io/region
+    - topology.kubernetes.io/zone
+  cephFS:
+    kernelMountOptions: readdir_max_bytes=1048576,norbytes
+    fuseMountOptions: debug
```
nb-ohad, line 329: I am sure we agreed that this value does not fit here. As all of the other values are dictated on Ceph while these ones are admin-provided and admin-managed. The value's life cycle and reason to change is completely different. 

 ```suggestion
```

File: docs/design/operator.md [hunk 28]
```diff
@@ -0,0 +1,414 @@
...
+```
+
+### CephCSICephConfig CRD
+
+Stores connection and configuration details to a Ceph cluster. Provides common
+connection and configuration information for multiple CSI drivers
+
+```yaml
+---
+kind: CephCSICephConfig
```
nb-ohad, line 312: I beleive we agreed to call this CR `CephCSICephCluster`

docs/design/operator.md
```diff
@@ -0,0 +1,414 @@
...
+    kernelMountOptions: ms_mode=secure
+status:
+  phase: Failed
+  message: Failed to create cephfs csi driver
+  reason: csi driver with same name already exists in the cluster
+```
+
+### CephCSICephConfig CRD
+
+Stores connection and configuration details to a Ceph cluster. Provides common
```
nb-ohad, line 307: This CR does not offer a configuration point, the values in the spec are always derived from the Ceph cluster itself so we should not mention "Configuration" in the description. 

```suggestion
Stores connection details for a single Ceph cluster. Provides common
```

docs/design/operator.md
```diff
@@ -0,0 +1,414 @@
...
+status:
+  phase: Succeeded
+  message: successfully linked to CephCSICephConnection
+```
+
+### CephCSIClusterRecovery
+
+The CephCSIClusterRecovery CR contains the local and the remote cephcluster
+configuration which will help cephcsi to identify the blockpool alias to
+consume the volume from when doing CSI operation.
```
nb-ohad, line 367: Let's completely revise this description based on my comment on the requested change of this CR kind/name

nb-ohad (changes requested):

docs/design/operator.md
Madhu-1: agree but no harm in keeping it as well.

docs/design/operator.md
Madhu-1: Lets discuss about pros and cons, this was based on the last few discussions.

docs/design/operator.md
Madhu-1: Yes thats correct, this is coved in common section

docs/design/operator.md
Madhu-1: i was referring to know API's for it https://github.com/kubernetes/api/blob/664bdd58a30d9e428a31e7e77713a26424025722/core/v1/types.go#L414-L425.

docs/design/operator.md
Madhu-1: for me it doesn't make sense to allow running two rbd csi drivers in the same namespace. if you don't agree with this restriction we can talk about it.

docs/design/operator.md
```diff
@@ -0,0 +1,414 @@
...
+```yaml
+---
+kind: CephCSIDriver
+apiVersion: csi.ceph.io/v1alpha1
+metadata:
+  name: "<prefix>.cephfs.csi.ceph.com"
+  namespace: operator-namespace
+spec:
+  deploymentName: csi-cephfsplugin-provisioner
+  daemonsetName: csi-cephfsplugin
```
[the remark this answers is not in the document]
Madhu-1: This is still under discussion we need to check how to give control to the user/admin for over taking existing resouces

docs/design/operator.md
Madhu-1: i am more inclined to keep it still as Recovery as we are trying to recover from a state that cannot be changed in the current cluster am open to alternative as well. Mapping might not be the right name

docs/design/operator.md
Madhu-1: That make sense, changed it back to name.

docs/design/operator.md
```diff
@@ -0,0 +1,407 @@
...
+```
+
+In this updated diagram:
+
+- **CRD Changes**: Represents changes made to Custom Resource Definitions,
+  which trigger actions in the operator.
+- **Operator**: Listens for changes in CRDs and initiates the installation and
+  configuration of CSI drivers.
+- **Configure CephFS, NFS, RBD**: Actions performed by the operator to install
+  and configure the respective CSI drivers based on the CRD changes.
```
BlaineEXE, line 67: This architecture is one-way only. `CRD changes` -> `Configure drivers`. I know that CSI drivers are usually pretty simple and can usually just be a deployment and daemonset, so maybe this really is the architecture. But the operator pattern follows control theory, where system states also result in the controller taking actions. 

Are there any other :
- are there ConfigMaps that cause reconciliations?
- does the controller respond to any other events like nodes going online/offline, or getting new labels applied?


I feel like this is a bigger question: is this operator taking on the node loss fencing behavior? https://github.com/rook/rook/blob/682d639532534068a6154664119e10f0016f547d/design/ceph/node-loss-rbd-cephfs.md

I feel that, since the node fencing is both Kube-specific and involves CSI-Addons CRDs, it makes sense that the Ceph-CSI driver would be the right controller for this, so that non-Rook users could benefit from the fencing that RBD and CephFS need to function well in failure conditions.

docs/design/operator.md
BlaineEXE: I usually see phase/message/reason in resource Conditions, but it seems to me that Conditions have been going out of vogue. I think this is probable because they are incredibly hard to deal with in code and user scripts compared to having typed fields for status.

Many of the recent sig-storage KEPs use a `someResourceReady: bool` (e.g., `bucketReady: bool`, `snapshotReady: bool`) as well, so that users can know with a boolean whether their resource is ready.

Personally, I don't have a preference between `message` and `reason`. Both are alphabetically close to `phase` which should mean users see the outputs near each other as status items are added.

But I would suggest some sort of `xReady` bool that indicates when config is ready.

BlaineEXE (commented):
I haven't gone full through this yet. I think the biggest question I have that we haven't talked about is what to do with the node fencing controller. IMO, this operator would be the best home for it for users.

docs/design/operator.md
```diff
@@ -0,0 +1,407 @@
...
+## Assumptions and Dependencies
+
+Kubernetes cluster is available and operational. Ceph cluster is set up and
+accessible from the Kubernetes cluster. Users are familiar with Kubernetes and
+Ceph concepts. CRD and operator versions are compatible with the Kubernetes
+version in use.
+
+## General Constraints
+
+Resource limitations of the Kubernetes cluster (CPU, memory). Network latency
```
travisn, line 28: How are these constraints specific to the Ceph-CSI operator? They seem applicable to anything in K8s. 

I'd suggest to use a bulleted list so each separate constraint is clear, and what the design point is for that constraint.

docs/design/operator.md
```diff
@@ -0,0 +1,407 @@
...
+
+## Goals and Guidelines
+
+- Automate the deployment and lifecycle for Ceph-CSI drivers.
+- Provide a consistent and flexible configuration mechanism using CRDs.
+- Enable namespace-scoped configurations for different Ceph-CSI drivers.
+- Ensure high availability and scalability of the operator.
+- Support multiple deployment methods for the operator (single-namespace,
+  multi-namespace, all-namespaces).
+- Support multiple operator deployment in a single cluster when running in a
```
travisn, line 41: 1. What is the difference in behavior between the single-namespace and multi-namespace modes? (Maybe it's listed later in the doc?) 
2. Will the default/recommended mode be single-namespaced? The single-namespaced mode seems the same pattern we have by default in Rook today.

docs/design/operator.md
```diff
@@ -0,0 +1,407 @@
...
+```mermaid
+graph TD
+    A[CRD Changes] --> B[Operator]
+    B -->  D[Configure CephFS CSI Driver]
+    B -->  E[Configure RBD CSI Driver]
+    B -->  F[Configure NFS CSI Driver]
+
+```
+
+In this updated diagram:
```
travisn, line 60:
```suggestion
In this diagram:
```

docs/design/operator.md
```diff
@@ -0,0 +1,407 @@
...
+where operator is running. The configurations are catagorized into 2 different
+parts
+
+- Operator configurations
+- Common configurations across all CSI driver
+
+```yaml
+---
+kind: CephCSIOperatorConfig
+apiVersion: csi.ceph.io/v1alpha1
```
travisn, line 85: Let's consider starting with v1:
- We are creating this based on a stable set of settings that we have been using in Rook or other deployments in production for some time. It's not a new unproven feature, it's already stable.
- If we have confidence in the settings and can plan to support upgrades and ensure backward compatibility, we should just start with v1.
- If we are expecting to immediately support the csi operator either upstream or downstream, we should *not* declare alpha. `v1alpha1` indicates it should not be used in production.
```suggestion
apiVersion: csi.ceph.io/v1
```

docs/design/operator.md
```diff
@@ -0,0 +1,407 @@
...
+metadata:
+  name: csioperatorconfig
+  namespace: operator-namespace
```
travisn, line 88: How about using the namespace `ceph-csi` in the examples?
```suggestion
  namespace: ceph-csi
```

docs/design/operator.md
```diff
@@ -0,0 +1,407 @@
...
+spec:
+  logLevel: 1
+  driverSpecDefaults:
+    logging:
+      logLevel: 5
+      maxfiles: 5
+      maxLogSize: 10M
+    clusterName: 5c63ad7e-74fe-4724-a511-4ccdc560da56
```
travisn, line 96: Is this the ceph fsid?
```suggestion
    cephClusterID: 5c63ad7e-74fe-4724-a511-4ccdc560da56
```

docs/design/operator.md
```diff
@@ -0,0 +1,407 @@
...
+            cpu: '100'
+            memory: '250'
+        plugin:
+          limits:
+            cpu: '200'
+            memory: '500'
+          requests:
+            cpu: '100'
+            memory: '250'
+      volumes:
```
travisn, line 149: The volumes need to be configurable? Looking at the current csi templates, I only see the kubelet dir as being configurable in the csi driver volumes, though I didn't look very deeply.

docs/design/operator.md
```diff
@@ -0,0 +1,407 @@
...
+            cpu: '100'
+            memory: '250'
+    liveness:
+      metricsPort: 8000
+    leaderElection:
+      leaseDuration: 100
+      renewDeadline: 100
+      retryPeriod: 10
+    deployCSIAddons: true
+    cephFsKernelClient: true
```
travisn, line 238: Shouldn't cephfs settings only be on a cephfs-specific setting? I thought the settings in this CRD should apply to all drivers.

docs/design/operator.md
travisn: I also prefer the restriction of a single driver instance per namespace. It matches the pattern we have for Rook also that is a single ceph cluster per namespace. It has helped keep some design aspects simpler throughout the project.

docs/design/operator.md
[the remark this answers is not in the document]
travisn: Why does the admin need to configure the daemonset or deployment name? This is one of the simplifications of only allowing one driver instance per namespace, we can just hard-code a name.

docs/design/operator.md
```diff
@@ -0,0 +1,407 @@
...
+---
+kind: CephCSIDriver
+apiVersion: csi.ceph.io/v1alpha1
+metadata:
+  name: "<prefix>.<driver_type>.csi.ceph.com"
+  namespace: operator-namespace
+spec:
+  deploymentName: csi-cephfsplugin-provisioner
+  daemonsetName: csi-cephfsplugin
+  Spec:
```
travisn, line 264: In the implementation, the Go types be shared between the CephCSIDriver CRD and the CephCSIOperatorConfig CRD, right? Then the common structs can be extended with the driver-specific settings.

docs/design/operator.md
```diff
@@ -0,0 +1,407 @@
...
+    fsGroupPolicy: File
+    encryption:
+      configMapRef:
+        name: encryption-config-map-name
+    plugin:
+      priorityClassName: system-node-critical
+      updateStrategy:
```
travisn, line 271: Why do we need to expose the updateStrategy configuration? Can't we just determine the settings we want to support during upgrades? We don't have this configuration exposed today, right?

docs/design/operator.md
```diff
@@ -0,0 +1,407 @@
...
+  cephFsKernelMountOptions: ms_mode=secure
+status:
+  phase: Failed
+  message: Failed to create cephfs csi driver
+  reason: csi driver with same name already exists in the cluster
+```
+
+### CephCSICephConfig CRD
+
+Stores connection and configuration details for a single single Ceph cluster.
```
travisn, line 303:
```suggestion
Stores connection and configuration details for a single Ceph cluster.
```

docs/design/operator.md
```diff
@@ -0,0 +1,407 @@
...
+Provides common connection and configuration information for multiple CSI
+drivers
+
```
travisn, line 306: Can you describe the flow for this CR? Is it something like this?
1. Rook creates this CR when it reconciles a CephCluster
2. Ceph-CSI operator evaluates the CR and creates/updates the configmap and secrets used by the csi drivers.

docs/design/operator.md
```diff
@@ -0,0 +1,407 @@
...
+  ReadAffinity:
+    crushLocationLabels:
+    - kubernetes.io/hostname
+    - topology.kubernetes.io/region
+    - topology.kubernetes.io/zone
+  cephFS:
+    kernelMountOptions: readdir_max_bytes=1048576,norbytes
+    fuseMountOptions: debug
+  rbd:
+    mirrorDaemonCount: 2
```
travisn, line 328: Rook controls the number of rbd mirror daemons. What does the CSI driver control with this setting?

docs/design/operator.md
```diff
@@ -0,0 +1,407 @@
...
+  cephConfig: |-
```
travisn, line 329: Consider implementing the ceph config similar to Rook's as seen in [this example](https://github.com/rook/rook/blob/master/deploy/examples/cluster-test.yaml#L49-L55)

docs/design/operator.md
```diff
@@ -0,0 +1,407 @@
...
+    auth_cluster_required = none
+    auth_service_required = none
+    auth_client_required = none
+    rbd_validate_pool = false
+status: {}
+```
+
+### CephCSIConfig CRD
+
+Contains details about CephFS subvolume group or RADOS namespace and a
```
travisn, line 340: Does this CRD replace the need to create a hash from the subvolume group name or rados namespace and add it to the storage class? Or how is it related?

docs/design/operator.md
```diff
@@ -0,0 +1,407 @@
...
+---
+kind: CephCSIConfig
+apiVersion: csi.ceph.io/v1alpha1
+metadata:
+  name: storage
+  namespace: operator-namespace
+spec:
+  cephClusterRef:
+    name: ceph-cluster-1
+  subvolumeGroup: csi
```
travisn, line 353: Several questions:
- What if you have multiple subvolumegroups or rados namespaces?
- Why are subvolume groups and rados namespaces both defined together? They are very different resources for different drivers.
- This CRD seems too generic and doesn't have a specific purpose. Even the CRD name "Config" indicates maybe it's just miscellaneous settings?

docs/design/operator.md
```diff
@@ -0,0 +1,407 @@
...
+
+### CephCSIConfigRecovery
+
+The CephCSIConfigRecovery CR contains the local and the remote cephcluster
+configuration which will help cephcsi to identify the blockpool alias to
+consume the volume from when doing CSI operation.
+
+```yaml
+---
+kind: CephCSIConfigRecovery
```
travisn, line 368: Is this a new concept for the csi driver? Or how is it specified in the driver today? I remember discussions about this feature, but don't recall if it's already there.

docs/design/operator.md
```diff
@@ -0,0 +1,407 @@
...
+      poolID: 2
+  - local:
+      cephCSIConfigName: remote1-cephCSICluster-name
+      poolID: 2
+    remote:
+      cephCSIConfigName: remote2-cephCSICluster-name
+      poolID: 3
+```
+
+## Glossary
```
travisn, line 389: IMO no need for the glossary. Anyone reading the doc should already understand basic K8s and Ceph concepts, or refer to their glossaries.

travisn (changes requested):

docs/design/operator.md
Rakshith-R: I think it should refer to this.

https://github.com/ceph/ceph-csi/blob/devel/cmd/cephcsi.go#L71 

https://github.com/rook/rook/blob/8dc88c39b105c79f8f5678f83dd23414411daa40/deploy/examples/operator.yaml#L59-L62

should this be overrideable at CephCSIDriver level ?

docs/design/operator.md
```diff
@@ -0,0 +1,407 @@
...
+  configuration of CSI drivers.
+- **Configure CephFS, NFS, RBD**: Actions performed by the operator to install
+  and configure the respective CSI drivers based on the CRD changes.
+
+## CRD's for ceph-csi-operator
+
+### CephCSIOperatorConfig CRD
+
+Manages operator-level configurations and offers a place to overwrite settings
+for CSI drivers. This CRD is a namespace-scoped CRD and a single CR named
```
Rakshith-R, line 74: These configurations override or act as default for all CephCSIDrivers in different ns/type ?

docs/design/operator.md
Madhu-1: >are there ConfigMaps that cause reconciliations?

There will be 2 controllers in cephcsi-operator
* One which takes care of deployment
* The one Reconciles all the configuration and update the configmap which is already mounted to the csi pods.

Good thought but I haven't thought about making csi-addons activities as part of the core operator we can decide on that where the operator needs to create extra CR but we can take this in the next iteration not in the first phase.

docs/design/operator.md
Madhu-1: As the name suggests the operator will watch for changes in single namespace and multi-namespace modes,
The suggestion completely depends on the way how admin wants to install and configure csi drivers how many csi drivers they wants to starts and how they want to manage it.

docs/design/operator.md
Madhu-1: i used operator-namespace as example to provide more clarity that we can have the CR's created in the operator-namespace, let me know if you still suggest to change it to ceph-csi.

docs/design/operator.md
Madhu-1: This refers to the clusterName where cephcsi driver is running (typically openshift clusterID or some arbitrary kubernetes cluster name admin sets.

docs/design/operator.md
Madhu-1: Yes we do have this flexibility in Rook as very based on different environments, sometime admin needs additional mounts to the csi driver. https://github.com/rook/rook/blob/8dc88c39b105c79f8f5678f83dd23414411daa40/deploy/examples/operator.yaml#L168-L196

docs/design/operator.md
Madhu-1: we have two options either we can have sections for each driver or we don't need extra section we can go with driver_type_configuration like that. as we will have very minimal configuration and sometimes a single configuration can be applicable for 2 drivers as well, for that reason we kept this way.

docs/design/operator.md
Madhu-1: Yes thats correct

docs/design/operator.md
Madhu-1: We have this configuration exposed today, thats the reason to make it configurable for user.

docs/design/operator.md
Madhu-1: yes thats correct but there wont be any secrets only configmap.

docs/design/operator.md
Madhu-1: This is required for cephcsi, this was added as part of https://github.com/ceph/ceph-csi/issues/4312 which is used to check active users of a single RBD image.

docs/design/operator.md
Madhu-1: The name will be the hash what Rook generates today for subvolumegroup or radosnamespace and same will be added to the storageclass, Rook need to create this new CR when there is new subvolumegroup/radosnamespace  created

docs/design/operator.md
Madhu-1: >What if you have multiple subvolumegroups or rados namespaces?

The user needs to create multiple CRs, its one CR represents either subvolumegroup or radosnamespace.

>Why are subvolume groups and rados namespaces both defined together? They are very different resources for different drivers.

This is a generic CRD and even though they are specific to csi driver, we are keeping it generic because the subvolume group applies to cephfs and nfs not for rbd.

This CRD seems too generic and doesn't have a specific purpose. Even the CRD name "Config" indicates maybe it's just miscellaneous settings?

The idea of keeping it generic which helps in extending it in future instead of adding new CRD's and Config indicates that this is a configuration to the cephcsi driver it can be generic and it can be driver-specific.

docs/design/operator.md
Madhu-1: No this is an existing configmap Rook is created today for the RBDPoolMapping. https://github.com/rook/rook/blob/8dc88c39b105c79f8f5678f83dd23414411daa40/pkg/operator/ceph/csi/peermap/config.go#L51-L54

docs/design/operator.md
Madhu-1: This is an override for the defaults which come as part of the go code within the operator and also this can be the default for all the csidriver this operator is installing and each csi driver can override the default present in this CR as well.

docs/design/operator.md
Madhu-1: Yes this is true am still not clear should we go with a bool value to indicate that the driver is running successfully after deployment where i need to ensure that the csi driver is able to serve the request for the user or not as it involves multiple check which are not exposed to the consumer yet. but i would like to keep this an option which we need explore in future, @BlaineEXE i hope that should be fine as well.

docs/design/operator.md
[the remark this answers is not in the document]
Madhu-1: This was added to allow taking ownership of the existing csi deployments but for now am dropping which we can take care of it later as well.

docs/design/operator.md
```diff
@@ -0,0 +1,383 @@
...
+  cephClusterRef:
+    name: ceph-cluster-1
+  subvolumeGroup: csi
+  radosNamespace: rados-test
+status:
+  phase: Succeeded
+  message: successfully linked to CephClusterRef
+```
+
+### CephCSIConfigRecovery
```
subhamkrai, line 352:
```suggestion
### CephCSIRecoveryConfig
```

in some place I see `CephCSI*Config` and in some place it is `CephCSIConfig*`. Ex: `CephCSICephConfig` and `CephCSIConfigRecovery`

docs/design/operator.md
```diff
@@ -0,0 +1,383 @@
...
+  cephFsKernelMountOptions: ms_mode=secure
+status:
+  phase: Failed
+  message: Failed to create cephfs csi driver
+  reason: csi driver with same name already exists in the cluster
+```
+
+### CephCSICephConfig CRD
+
+Stores connection and configuration details for a single Ceph cluster.
```
parth-gr, line 295: I guess we should mention in all the crds docs which resource is cluster scoped and which is not

Madhu-1: none of the CRS's are cluster scoped and all are namespace scoped, in the CR you can see the namespace field, that ensure that a CR is namespace scoped

docs/design/operator.md
```diff
@@ -0,0 +1,383 @@
...
+---
+kind: CephCSIOperatorConfig
+apiVersion: csi.ceph.io/v1
+metadata:
+  name: csioperatorconfig
+  namespace: operator-namespace
+spec:
+  logLevel: 1
+  driverSpecDefaults:
+    logging:
```
parth-gr, line 87: Does this logging setting go to new CRDS, (all the csi containers)?

I think we can have a separate section
```
logCollector:
    enabled: true
    periodicity: daily # one of: hourly, daily, weekly, monthly
    maxLogSize: 500M # SUFFIX may be 'M' or 'G'. Must be at least 1M.
```
We should keep it similar way that we have for ceph daemons

parth-gr (changes requested):

docs/design/operator.md
Madhu-1: This is a separate section for logging but we don't have any plan to have common keys that represent ceph as these are for sidecar and cephcsi logging we will try to keep the configuration fields similar to Linux/kubernetes log rotation. we can  change the fields later on once we have a solid design that you are working on related to the configurations.

docs/design/operator.md
travisn: All CRDs are implicitly configuration, so it feels redundant IMO to include "Config" in CRDs names. And based on the settings and its purpose, perhaps consider "PeerMapping" or "PoolMapping" in the name? Or at least the name "Recovery" is not clear to me.
```suggestion
### CephCSIPeerMapping
```

docs/design/operator.md
nb-ohad: @travisn a single CSI cluster configuration supports all driver types. This is the way Ceph-CSI was built. This new operator is designed to work on top of the established way the underlying CSI driver is configurable and as such each CSI cluster config needs to be able to specify options for all driver types simultaneously

Madhu-1: Adding to the above `SubvolumeGroup` and `RadosNamespace` becomes more ceph terminology and it's really an abstraction for us for where the storage is to be consumed from later we can extend that to new objects if ceph adds anything new instead of adding more CRD's to it.

docs/design/operator.md
Madhu-1: I changed to the `CephCSIConfigMapping` as its a Mapping between the configuration even inside the local cluster itself not a new configuration which always connects to the remote cluster, let me know if it need to be changed

docs/design/operator.md
Madhu-1: @travisn i choose to keep `operator-namespace` in the design doc which gives details on what was the design consideration but as you suggested with names that is something we will add it to the deployment section (in documentation) or the deployment yamls

docs/design/operator.md
Madhu-1: https://github.com/rook/rook/blob/8dc88c39b105c79f8f5678f83dd23414411daa40/deploy/examples/operator.yaml#L59-L61 This is some configuration even Rook is not detecting it today and user is adding it. In openshift the clusterID is something specific to openshift CRD and in kubernetes we don't have any API for clusterID and its something user can choose alternative names they want for this field.

docs/design/operator.md
Madhu-1: This is specific to multiple cephfs driver deployed by the single/multiple cephcsi operator, as this is operator config its global configuration for all the cephfs drivers its managing.

docs/design/operator.md
Madhu-1: Yes thats correct, this is something Rook/user needs to automate based on the changes in the CephRBDMirror CR.

docs/design/operator.md
Madhu-1: This is moved to different CR.

docs/design/operator.md
Madhu-1: closing as new issue is opened for it.

docs/design/operator.md
Madhu-1: Resolving as new issue is opened for it.

docs/design/operator.md
```diff
@@ -0,0 +1,382 @@
...
+  phase: Succeeded
+  reason: operator config successfully created
+```
+
+### CephCSIDriver CRD
+
+Manages the installation and configuration of CephFS, RBD, and NFS CSI drivers
+within namespaces. Allows customization of driver settings on a per-namespace
+basis. Best practise dictates only a single instance of RBD/NFS/CephFS drivers
+should be created in a single namespace, There is a neeed to deploy a second
```
SanjalKatiyar, line 245:
```suggestion
should be created in a single namespace, there is a need to deploy a second
```

docs/design/operator.md
nb-ohad: @travisn this CR offers a place to overwrite all default values, even if only applied to a single driver type. 
This is not a configuration for a driver and will not result in deployment of a driver

docs/design/operator.md
```diff
@@ -0,0 +1,382 @@
...
+status:
+  phase: Failed
+  reason: csi driver with same name already exists in the cluster
+```
+
+### CephCSICephConfig CRD
+
+Stores connection and configuration details for a single Ceph cluster.
+Provides common connection and configuration information for multiple CSI
+drivers
```
nb-ohad, line 296:
```suggestion
Stores connection and configuration details for a single Ceph cluster and
provide the information use by multiple CSI drivers
```

docs/design/operator.md
```diff
@@ -0,0 +1,382 @@
...
+      auth_service_required: none
+      auth_client_required: none
+      rbd_validate_pool: false
+status: {}
+```
+
+### CephCSIConfig CRD
+
+Contains details about CephFS subvolume group or RADOS namespace and a
+reference to CephCSIConfig to link storage configurations to Ceph cluster.
```
nb-ohad, line 328:
```suggestion
Contains details about CephFS, RBD, and NFS configuration to be used when communicating with Ceph. Include a reference to a CephCSICephCluster holding the connection information for the target Ceph cluster 
```

docs/design/operator.md
```diff
@@ -0,0 +1,382 @@
...
+status:
+  phase: Succeeded
+  reason: successfully linked to CephClusterRef
+```
+
+### CephCSIConfigMapping
+
+The CephCSIConfigMapping CR contains the local and the remote cephcluster
+configuration which will help cephcsi to identify the blockpool alias to
+consume the volume from when doing CSI operation.
```
nb-ohad, line 355:
```suggestion
The CephCSIConfigMapping CR contains a mapping between local and remote Ceph cluster
configurations. The information in this CR helps Ceph CSI identify peered blockpools that can help with
the correct management and consumption of volumes in backup and DR scenarios 
```

docs/design/operator.md
```diff
@@ -0,0 +1,382 @@
...
+
+## Goals and Guidelines
+
+- Automate the deployment and lifecycle for Ceph-CSI drivers.
+- Provide a consistent and flexible configuration mechanism using CRDs.
+- Enable namespace-scoped configurations for different Ceph-CSI drivers.
+- Ensure high availability and scalability of the operator.
+- Support multiple deployment modes for the operator:
+  - Single-namespace: the operator manages a single Ceph-CSI driver in the
+    current namespace where operator is deployed.
```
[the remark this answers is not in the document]
Madhu-1: yes and i think i dont need to explicitly need to mention about single namespace

docs/design/operator.md
```diff
@@ -0,0 +1,382 @@
...
+    leaderElection:
+      leaseDuration: 100
+      renewDeadline: 100
+      retryPeriod: 10
+    attachRequired: true
+  liveness:
+    metricsPort: 8000
+  deployCSIAddons: false
+  cephFsKernelClient: true
+  cephFsKernelMountOptions: ms_mode=secure
```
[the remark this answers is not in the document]
Madhu-1: am still not fine with one , we need to bring this under cephfs or filesystem section but for now, lets get this merged and bring that as a new Issue later on.

docs/design/operator.md
nb-ohad: @travisn This is one way to achieve it. I will evaluate all of the options when I write the structs

docs/design/operator.md
```diff
@@ -0,0 +1,386 @@
...
+- Support multiple deployment modes for the operator:
+  - Own-namespace: the operator manages a single Ceph-CSI driver in the
+    current namespace where operator is deployed.
+  - Multi-namespace: the operator manages Ceph-CSI drivers in user-specified
+    namespaces.
+  - All-namespaces: the operator manages Ceph-CSI drivers in all namespaces.
+
+## Development Methods
+
+Use kubebuilder framework which is higly used in operator development ,
```
nb-ohad, line 41:
```suggestion
Use kubebuilder framework which is highly used in operator development,
```

docs/design/operator.md
```diff
@@ -0,0 +1,386 @@
...
+
+## CRDs for ceph-csi-operator
+
+### CephCSIOperatorConfig CRD
+
+Manages operator-level configurations and offers a place to overwrite settings
+for CSI drivers. This CRD is a namespace-scoped CRD and a single CR named
+ceph-csi-operator-config can be created by an admin inside the operator's
+namespace.
+The configurations are catagorized into 2 different types
```
nb-ohad, line 72:
```suggestion

The configurations are categorized into 2 different types
```

docs/design/operator.md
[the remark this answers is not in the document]
nb-ohad: ok

docs/design/operator.md
```diff
@@ -0,0 +1,386 @@
...
+  kernelMountOptions: ms_mode=secure
+status:
+  phase: Failed
+  reason: csi driver with same name already exists in the cluster
+```
+
+### CephCSICephConfig CRD
+
+Stores connection and configuration details for a single Ceph cluster and
+provide the information use by multiple CSI drivers
```
nb-ohad, line 298:
```suggestion
provide the information to be used by multiple CSI drivers
```

docs/design/operator.md
```diff
@@ -0,0 +1,386 @@
...
+status:
+  phase: Succeeded
+  reason: successfully linked to CephClusterRef
+```
+
+### CephCSIConfigMapping
+
+The CephCSIConfigMapping CR contains a mapping between local and remote Ceph
+cluster configurations. The information in this CR helps Ceph-CSI identify
+peered blockpools that can help with the correct management and consumption of
```
nb-ohad, line 358:
```suggestion
peered blockpools.  This can help with the correct management and consumption of
```

nb-ohad (changes requested):

nb-ohad:
LGTM

docs/design/operator.md
travisn: Ok understood now on why these settings are needed in this CRD. What about a section for the `cephfs` type, just to organize the settings? 
```suggestion
    cephfs:
      kernelClient: true
      kernelMountOptions: ms_mode=secure
```

docs/design/operator.md
travisn: What if this clusterName is not set? Since the setting today is not set by default for Rook upstream users I assume it's not required. It's only needed for some scenarios, right? Are those features disabled if not set?

docs/design/operator.md
```diff
@@ -0,0 +1,387 @@
...
+### CephCSIConfigMapping
+
+The CephCSIConfigMapping CR contains a mapping between local and remote Ceph
+cluster configurations. The information in this CR helps Ceph-CSI identify
+peered blockpools. This can help with the correct management and consumption of
+volumes in backup and DR scenarios
+
+```yaml
+---
+kind: CephCSIConfigMapping
```
travisn, line 364: I think I commented on this previously, but can't find it now or if there was a response... "Config" in the CRD name is redundant IMO. How about a name without "Config"? For example, CephCSIPeerMapping, CephCSIPoolMapping, CephCSIClusterMapping, or even just CephCSIMapping
```suggestion
kind: CephCSIPeerMapping
```

docs/design/operator.md
```diff
@@ -0,0 +1,387 @@
...
+namespace.
+
+The configurations are catagorized into 2 different types
+
+- Operator configurations
+- Default configuration overwrites applied to all Ceph CSI drivers
+
+```yaml
+---
+kind: CephCSIOperatorConfig
```
travisn, line 80: Suggestion to remove "Config" from all CRD names. Almost all the CRD names in the design include it and it just is redundant IMO. All CRDs contain configuration settings.

travisn (changes requested):

docs/design/operator.md
nb-ohad: We deliberated toward and away from a dedicated section. There are good reasons to do it and good reasons to not. 

But consider the following: For a single driver, a section does not make sense as it can only be of one type - while a section suggests that you should be able to specify many.

Given that the above was the leading reason we opted out of a section for the driver CR. We decided to match the operator's config CR with an identical structure, to simplify the coding (reuse the same struct) and for familiarity and consistency.

docs/design/operator.md
nb-ohad: Here config is a subject, not an adjunctive. So it makes sense to keep it.

travisn: My point is that everything in every CRD is config, why does it need to be part of the CRD name?

BlaineEXE: I can see Travis' point here. It's obvious that it's a subject, but it seems redundant in the sense that all k8s resources contain config (or spec, as k8s would call it). Is there a more clear name for this/these?

CephCSIPeerMapping does seem to be more descriptive and explanatory to me, based on the description:

> The CephCSIConfigMapping CR contains a mapping between local and remote Ceph
cluster configurations. The information in this CR helps Ceph-CSI identify
peered blockpools. This can help with the correct management and consumption of
volumes in backup and DR scenarios

IMO, even "mapping" is a bit confusing here. I feel that "mapping" may be too unclear, or just redundant.

IMO, CephCSIPeer is a sufficient name to indicate that
1. This configures Ceph-CSI
2. This configures Ceph-CSI with "peer" information

docs/design/operator.md
```diff
@@ -0,0 +1,387 @@
...
+  second instance of a driver of a similar type, it should be deployed in a
+  different namespace.
+- Each CSI driver must have a unique name across all namespaces.
+
+```yaml
+---
+kind: CephCSIDriver
+apiVersion: csi.ceph.io/v1alpha1
+metadata:
+  name: "<prefix>.<driver_type>.csi.ceph.com"
```
BlaineEXE, line 258: I have some concerns about using the CR name field to encode config options like this.

On one hand, I understand that this avoids certain kinds of complexity.

On the other hand, this feels like it would be hard for users to understand.
Do I have to provide a prefix? Does it have to be unique? Can I **not** provide a prefix if I don't want to?

What are the available driver types I can use? This is to say that the only indication I had from this example that it was a cephfs-specific driver was from `labels.app=cephfs-plugin`.

Considering how users will view this and how this will be documented, I think this will be less clear and that this is the kind of thing I see users get wrong often. 

I would suggest something like this as an alternative:
1. metadata.name can be whatever the user wants, with a reasonable suggestion being something like, "platform-cephfs-driver"
2. `prefix` becomes a required spec field
3. `driverType` becomes a required spec field with an `enum` set of options -- this makes it more clear (without having to have inside-knowledge, what the driver type is)
4. the design doc should specify what validations the operator will take to ensure that the resultant driver name is unique in the k8s cluster
5. to support usability/user feedback, add `status.driverName` = `<prefix>.<driverType>.csi.ceph.com` so that users have feedback to understand what driver name to use in StorageClasses.

docs/design/operator.md
```diff
@@ -0,0 +1,387 @@
...
+```
+
+### CephCSICephConfig CRD
+
+Stores connection and configuration details for a single Ceph cluster and
+provide the information to be used by multiple CSI drivers.
+
+```yaml
+---
+kind: CephCSICephCluster
```
BlaineEXE, line 303: CephCSICephConfig and CephCSICephCluster don't match. Seems like something was updated and one was forgotten -- probably CephCSICephCluster is the correct one?

docs/design/operator.md
```diff
@@ -0,0 +1,387 @@
...
+  cephFS:
+    subvolumeGroup: csi
+    kernelMountOptions: readdir_max_bytes=1048576,norbytes
+    fuseMountOptions: debug
+  rbd:
+    radosNamespace: rados-test
+status:
+  phase: Succeeded
+  reason: successfully linked to CephClusterRef
+```
```
BlaineEXE, line 353: Why is this not part of CephCSIDriver CRD?

BlaineEXE (changes requested):
As a general statement, I don't see any details in the design that explain how things work from an admin- or user-based view. I think admin/user workflows would help understand how things fit together, rather than just viewing individual pieces of the design.

At minimum, I think it would be very helpful to explain some details about fields that are important for cross-referencing between admin and user resources. And also, what are the limitations, and what are important field validations that must be done to prevent multi-user collisions.

docs/design/operator.md
```diff
@@ -0,0 +1,387 @@
...
+- Automate the deployment and lifecycle for Ceph-CSI drivers.
+- Provide a consistent and flexible configuration mechanism using CRDs.
+- Enable namespace-scoped configurations for different Ceph-CSI drivers.
+- Ensure high availability and scalability of the operator.
+- Support multiple deployment modes for the operator:
+  - Own-namespace: the operator manages a single Ceph-CSI driver in the current
+    namespace where operator is deployed.
+  - Multi-namespace: the operator manages Ceph-CSI drivers in user-specified
+    namespaces.
+  - All-namespaces: the operator manages Ceph-CSI drivers in all namespaces.
```
BlaineEXE, line 37: Does this mode auto-deploy drivers in all namespaces, or does it merely allow them to be deployed in all namespaces?

docs/design/operator.md
```diff
@@ -0,0 +1,387 @@
...
+
+## Development Methods
+
+Use kubebuilder framework which is highly used in operator development,
+following Kubernetes best practices for controllers and operators
+
+## Architecture
```
BlaineEXE, line 44: I feel like it would be best to include node fencing here, instead of treating it as a follow-on design: https://github.com/ceph/ceph-csi-operator/issues/3#issuecomment-2135845320

The node fencing addition changes the architecture significantly by making the operator no longer be a one-way graph. Instead, the system must also respond to other system changes.

docs/design/operator.md
BlaineEXE, line 37: Where is the API control for this? I can't find it in the doc.

docs/design/operator.md
```diff
@@ -0,0 +1,387 @@
...
+```yaml
+---
+kind: CephCSIOperatorConfig
+apiVersion: csi.ceph.io/v1alpha1
+metadata:
+  name: ceph-csi-operator-config
+  namespace: <operator-namespace>
+spec:
+  logLevel: 1
+  driverSpecDefaults:
```
BlaineEXE, line 87: IMO, the deployment mode (the operator's 'watch' scope) control makes the most sense here


something like...
```yaml
spec:
  watchNamespaces: # if empty, only watch current ns, ['*'] means watch all
    - ns1
    - ns2
    - etc
```

docs/design/operator.md
BlaineEXE, line 364: Different question related to peer mapping: why is this a separate CRD and not part of CephCSIDriver? Shouldn't each driver know what its peers are? What is the user story for this being separate versus included in CephCSIDriver?

I feel like this CRD seems strange to be a separate one in the multi-tenant case where a cluster admin wants to allow other users (or sub-admins) to spin up their own CSI drivers.

BlaineEXE (changes requested):
When I review this design doc from the standpoint of an admin who is allowing multi-tenant ability to deploy Ceph-CSI drivers, I start to become confused about how things fit together.

I think it will be helpful to explicitly explain how much multi-tenancy is intended to be designed into the architecture. What should be allowed? What should be explicitly disallowed? Will those details change in the future?

docs/design/operator.md
Madhu-1: As the CR's are namespace scoped resources and the operator if its deployed in this more it will manage all the csi driver which are created in their respective namespace. This is the pattern followed by many operator, do you think we need to add more details here?

Madhu-1: This doc doesnt talk about the API control for this, that going to come up in followup PR's.

docs/design/operator.md
Madhu-1: @BlaineEXE At least I don't have plans to include the fencing operation as part of the initial design, and we are still not sure about tying the CephCSI operator with the CSI-addons functionality as of today. Yes, as you mentioned, things might require changes based on the new features, and this architecture diagram needs to be updated based on the new additions.

docs/design/operator.md
Madhu-1: sounds good, am opening a new issue for it. https://github.com/ceph/ceph-csi-operator/issues/7

docs/design/operator.md
Madhu-1: Yes you are correct, This is not a mandatory parameter, if its not set we will disable the feature in backend

docs/design/operator.md
Madhu-1: @travisn if you are still not okay with this, am tracking similar one here https://github.com/ceph/ceph-csi-operator/issues/6 we can discuss on that

docs/design/operator.md
Madhu-1: >Do I have to provide a prefix? Does it have to be unique? Can I not provide a prefix if I don't want to?


This is not always required but that can be set but the format need to be followed.


> What are the available driver types I can use? This is to say that the only indication I had from this example that it was a cephfs-specific driver was from labels.app=cephfs-plugin.

This is just an example in the design doc and more specific CR will be provided when we have proper deployment guide.

>Considering how users will view this and how this will be documented, I think this will be less clear and that this is the kind of thing I see users get wrong often.

>// The name MUST follow domain name notation format
  // (https://tools.ietf.org/html/rfc1035#section-2.3.1). It SHOULD
  // include the plugin's host company name and the plugin name,
  // to minimize the possibility of collisions. It MUST be 63
  // characters or less, beginning and ending with an alphanumeric
  // character ([a-z0-9A-Z]) with dashes (-), dots (.), and
  // alphanumerics between. This field is REQUIRED.

we already through a couple of options before putting up the design doc with the same idea you have presented. we wanted to keep this CR the way how the csidriver looks today, the should be the unique name what the user wants the driver to the named. the CSI driver has some hard contstrain with the naming convention i mentioned above, Instead of introducing new field

with prefix and type in the spec it gives user an option to create duplicate csi driver easily with some which we are really trying to avoid to an extend by moving it to the name as the name of the CR is the name of the csi driver.

@nb-ohad anything to add?

docs/design/operator.md
Madhu-1: Yes thats is the correct one, this represents the ceph cluster details that's why its named as the `CephCSICephCluster`

docs/design/operator.md
Madhu-1: thats because a single cephcsi driver name have multiple ceph clusters created and its not required when we are starting a cephcsi driver, these details can be added later any point of time.

docs/design/operator.md
Madhu-1: IMO `CephCSIConfigMapping` is the right name for it because this collates the relationship of two things here only is the ceph pool mapping and also the csi configuration Name `cephCSIConfigName` , and its named a `CephCSIConfigMapping` because its a Mapping to the `CephCSIConfigMapping`, 


The usecase of this is that there are PVC exists with pool A but for some reason pool A was recover but it got a new pool ID for that cases, the user can still creating this CR to allow cephcsi to work with existing PVC's (this is totally low level on how cephcsi work and how we can recover few things), Currently we don't have support for this in Rook but we have support for this in cephcsi. This CR is really targeting to cover multiple users (cases).

>CephCSIPeerMapping does seem to be more descriptive and explanatory to me, based on the description:

The CephCSIConfigMapping CR contains a mapping between local and remote Ceph
cluster configurations. The information in this CR helps Ceph-CSI identify
peered blockpools. This can help with the correct management and consumption of
volumes in backup and DR scenarios

This was one more details that missing in the description but it was present in the CR field. i added more details to it i hope now this makes sense.

Madhu-1: This is the Peer information related to the configuration which Rook is generating here https://github.com/rook/rook/blob/f5ca09e38582d8658491e620e302c174ea11020f/pkg/operator/ceph/csi/peermap/config.go#L51-L110. This doesn't required when cephcsi driver is installed or in most of the normal cases, this is required only for recovery cases, that why its a different CRD.

nb-ohad: The information inside this CR maps between entities of type `CephCSIConfig`. This is why it is natural to call it `CephCSIConfig + Mapping = CephCSIConfigMaping` 

The reason we call the original entity `CephCSIConfig` is because it represents a configuration entry for CSI, we cannot just drop the config part. If we do so it becomes `CephCSI` which does not make sense. 

Regarding the `CephCSIOperatorConfig` name: A config map of a similar name is a common practice for operators (`<operator-name>-operator-config`), and is even provided by default. We just elevated the concept into a CR

docs/design/operator.md
nb-ohad: @BlaineEXE we do understand this is not a perfect solution but a decision based on "the best of two evils" approach. The requirements are as follows:

All driver names have to be unique! and conflict needs to be caught at creation time, not at reconciliation time. The name will be carried to the actual driver name as we want a 1:1 match between the client-exposed name at the operator layer and the name in the CSI implementation layer.

We dropped the type as a field in the spec for 2 reasons:
- I can create a CR that is called `my-rbd-driver` and have it of type `cephfs`
- Even if we agree that it is a user mistake, we cannot force the requirement to match the CSI driver implementation name to be `my-rbd-driver` because of CSI name restrictions 

On the other hand, if we encode the type in the name we can solve all the issues with some simple cel rules which gives the user a pretty good experience when it come to explanting the limitation when creation fails.

docs/design/operator.md
nb-ohad: All namespaces deploy the driver in a single dedicated namespace while allowing it to reconcile CRs across the entire cluster. To more exact, it allow the operator to be deployed as part of an operator group marked as an all namespaces operator group

docs/design/operator.md
nb-ohad: No, we cannot do that. The deployment control is part of the OLM spec and needs to be identified there in order for this to work. I Don't want us to create an operator that has internally code to reconcile OLM entities.

docs/design/operator.md
nb-ohad: 1. This CRD covers cross-driver concerns. 
2. It is a solution to a problem that is not inherited in CSI or its implementation. The problem only manifests itself when we add a layer of cross-cluster replication/mapping of volumes, which today only manifests as a DR-related concern. 

The above two points, and most specifically the second point, highlight that this is not a core concern or a core feature. With that in mind, it made sense to separate into this concern into its own CRD that can evolve on its own and does not impact the evolution of the core CRD as this operator matures

docs/design/operator.md
Madhu-1: Now i got it, it was a overlook from my side, fixed it.

docs/design/operator.md
travisn: It's a small detail, it's fine either way

docs/design/operator.md
travisn: I still am of the opinion that all of these would be better named without "Config", but let's take it as a separate discussion point and not block the design PR. Naming always takes time to settle. How about we open a separate issue for the discussion?

travisn (approved):

PR #16, opened 2024-06-20
 ci: add markdown lint

subhamkrai:
adding markdown lint CI check and also adding some small changes
to fix the CI

```
MD010: false # Hard tabs
MD012: false # Multiple consecutive blank lines
MD013: false # Line length
MD019: false # Multiple spaces after hash on atx style heading
MD021: false # Multiple spaces inside hashes on closed atx style heading
MD022: false # Headings should be surrounded by blank lines
MD024: false # Multiple headings with the same content
MD025: false # Multiple top-level headings in the same document
MD027: false # Multiple spaces after blockquote symbol
MD030: false # Spaces after list markers
MD031: false # Fenced code blocks should be surrounded by blank lines
MD032: false # Lists should be surrounded by blank lines
MD041: false # First line in a file should be a top-level heading
MD043: false # Required heading structure
MD045: false # Images should have alternate text (alt text)
MD053: false # Link and image reference definitions should be needed
```

### DIFF
--- a/.github/workflows/lint.yaml
+++ b/.github/workflows/lint.yaml
@@ -78,6 +78,18 @@ jobs:
       - name: govulncheck
         uses: golang/govulncheck-action@v1
 
+  markdownlint:
+    name: markdownlint
+    runs-on: ubuntu-latest
+    steps:
+    - uses: actions/checkout@v4
+      with:
+        fetch-depth: 1
+    - uses: DavidAnson/markdownlint-cli2-action@v16
+      with:
+        config: .markdownlint.yaml
+        globs: 'docs/*.md'
+
   shellcheck:
     name: Shellcheck
     runs-on: ubuntu-latest
--- a/.markdownlint.yaml
+++ b/.markdownlint.yaml
@@ -1,0 +1,16 @@
+MD010: false # Hard tabs
+MD012: false # Multiple consecutive blank lines
+MD013: false # Line length
+MD019: false # Multiple spaces after hash on atx style heading
+MD021: false # Multiple spaces inside hashes on closed atx style heading
+MD022: false # Headings should be surrounded by blank lines
+MD024: false # Multiple headings with the same content
+MD025: false # Multiple top-level headings in the same document
+MD027: false # Multiple spaces after blockquote symbol
+MD030: false # Spaces after list markers
+MD031: false # Fenced code blocks should be surrounded by blank lines
+MD032: false # Lists should be surrounded by blank lines
+MD041: false # First line in a file should be a top-level heading
+MD043: false # Required heading structure
+MD045: false # Images should have alternate text (alt text)
+MD053: false # Link and image reference definitions should be needed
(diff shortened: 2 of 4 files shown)


.github/workflows/lint.yaml
```diff
@@ -0,0 +1,86 @@
...
+        with:
+          fetch-depth: 0
+      - uses: actions/setup-go@v5
+        with:
+          go-version-file: go.mod
+      - name: golangci-lint
+        uses: golangci/golangci-lint-action@v6
+        with:
+          # Required: the version of golangci-lint is required and must be specified without patch version: we always use the latest patch version.
+          version: v1.55
```
leelavg, line 61: many times we observe below w/ golangci-lint:
1. versions in makefile & actions script quickly gets diverged (solvable)
2. we may use relatively latest linter on older codebase/release which might force us upgrading unrelated dependencies in respective branches (may need small discussion)

I know it's in draft and only leaving pointers atm, btw, description say `markdown lint` and content tries to lint more than that, working title may need to change by the end.

Madhu-1: we should have a new file like `build.env` where we will have the version mentioned and if its used in multiple place we just need to source and use that

Madhu-1:
@subhamkrai can you please update this PR, we have md files in the repo now.

nb-ohad:
@subhamkrai Can we have the list of agreed rules copied from the slack discussion here? (we can have the agreed list here as a comment referring to the offline discussion we had)

subhamkrai:
> @subhamkrai Can we have the list of agreed rules copied from the slack discussion here? (we can have the agreed list here as a comment referring to the offline discussion we had)



> @subhamkrai can you please update this PR, we have md files in the repo now.

I missed the ping, I'll check

subhamkrai:
@Madhu-1 let me know if the changes look good, I'll mark it ready for review or I also need to fix the CI to pass the linter

nb-ohad:
@subhamkrai the changes look good but applying the change results in CI failures. 
You will need to fix the issue to allow this PR to be merged

obnoxxx:
@subhamkrai,

just a hought:

Not sure what is the standard in this project, but wouldn't it be nice and convenient to add a make target for running markdown lint so that it can easily be invoked locally by developers
and this same make target would be invoked in the ci workflow?


@nb-ohad , @Madhu-1  - wdyt?

Madhu-1:
> @subhamkrai,
> 
> just a hought:
> 
> Not sure what is the standard in this project, but wouldn't it be nice and convenient to add a make target for running markdown lint so that it can easily be invoked locally by developers and this same make target would be invoked in the ci workflow?
> 
> @nb-ohad , @Madhu-1 - wdyt?

@obnoxxx yes that's correct, as its a separate configuration file for md rules, we can add a new target in makefile to run it locally so that , The developer can reproduce the issue related to linter locally instead of pushing the changes to the github always to see if the issue is resolved.

nb-ohad:
@Madhu-1 We can have the work the @obnoxxx is suggesting as a separate PR after this one merges

nb-ohad:
LGTM

Madhu-1 (approved):

PR #17, opened 2024-06-20
bundle: Implement changes for bundle generation

iamniting:
Signed-off-by: Nitin Goyal <nigoyal@redhat.com>

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -1,5 +1,16 @@
 # Image URL to use all building/pushing image targets
-IMG ?= controller:latest
+IMAGE_REGISTRY ?= quay.io
+REGISTRY_NAMESPACE ?= ocs-dev
+IMAGE_TAG ?= latest
+IMAGE_NAME ?= ceph-csi-operator
+BUNDLE_IMAGE_NAME ?= $(IMAGE_NAME)-bundle
+
+# IMG defines the image used for the operator.
+IMG ?= $(IMAGE_REGISTRY)/$(REGISTRY_NAMESPACE)/$(IMAGE_NAME):$(IMAGE_TAG)
+
+# BUNDLE_IMG defines the image used for the bundle.
+BUNDLE_IMG ?= $(IMAGE_REGISTRY)/$(REGISTRY_NAMESPACE)/$(BUNDLE_IMAGE_NAME):$(IMAGE_TAG)
+
 # ENVTEST_K8S_VERSION refers to the version of kubebuilder assets to be downloaded by envtest binary.
 ENVTEST_K8S_VERSION = 1.29.0
 
@@ -20,6 +31,25 @@ CONTAINER_TOOL ?= docker
 # Options are set to exit when a recipe line exits non-zero or a piped command fails.
 SHELL = /usr/bin/env bash -o pipefail
 .SHELLFLAGS = -ec
+
+# the PACKAGE_NAME is included in the bundle/CSV and is used in catalogsources
+# for operators (like OperatorHub.io). Products that include the ceph-csi-operator
+# bundle should use a different PACKAGE_NAME to prevent conflicts.
+PACKAGE_NAME ?= ceph-csi-operator
+
+# Creating the New CatalogSource requires publishing CSVs that replace one Operator,
+# but can skip several. This can be accomplished using the skipRange annotation:
+SKIP_RANGE ?=
+
+# Each CSV has a replaces parameter that indicates which Operator it replaces.
+# This builds a graph of CSVs that can be queried by OLM, and updates can be
+# shared between channels. Channels can be thought of as entry points into
+# the graph of updates:
+REPLACES ?=
+
+# The default version of the bundle (CSV) can be found in
+# config/manifests/bases/ceph-csi-operator.clusterserviceversion.yaml
+BUNDLE_VERSION ?= 0.0.1
 
 .PHONY: all
 all: build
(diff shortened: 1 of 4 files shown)


Makefile
```diff
@@ -120,6 +150,23 @@ build-installer: manifests generate kustomize ## Generate a consolidated YAML wi
...
 	$(KUSTOMIZE) build config/default > dist/install.yaml
 
+# generate the <package-name>.clusterserviceversion.yaml base
+gen-csv-base:
+	sed 's/@PACKAGE_NAME@/$(PACKAGE_NAME)/g;s/@SKIP_RANGE@/"$(SKIP_RANGE)"/g;s/@REPLACES@/"$(REPLACES)"/g' \
+	< config/manifests/bases/clusterserviceversion.yaml.in > config/manifests/bases/$(PACKAGE_NAME).clusterserviceversion.yaml
+
+.PHONY: bundle
+bundle: kustomize operator-sdk manifests gen-csv-base
+	$(KUSTOMIZE) build config/manifests | $(OPERATOR_SDK) generate bundle --manifests --metadata --package $(PACKAGE_NAME) --version $(BUNDLE_VERSION)
```
nb-ohad, line 160: Please do not assume we have operator SDK installed or even kustomize.

config/manifests/kustomization.yaml
nb-ohad, line 1: We don't need Kustomize to deliver OLM artifacts

Makefile
```diff
@@ -183,6 +232,16 @@ golangci-lint: $(GOLANGCI_LINT) ## Download golangci-lint locally if necessary.
...
 
+.PHONY: operator-sdk
+operator-sdk: ## Download operator-sdk locally.
+	@test -f $(OPERATOR_SDK) && echo "$(OPERATOR_SDK) already exists. Skipping download." && exit 0 ;\
+	echo "Downloading $(OPERATOR_SDK)" ;\
+        set -e ;\
+        mkdir -p $(dir $(OPERATOR_SDK)) ;\
+        OS=$(shell go env GOOS) && ARCH=$(shell go env GOARCH) && \
+        curl -sSLo $(OPERATOR_SDK) https://github.com/operator-framework/operator-sdk/releases/download/v${OPERATOR_SDK_VERSION}/operator-sdk_$${OS}_$${ARCH} ;\
+        chmod +x $(OPERATOR_SDK)
```
nb-ohad, line 243: We should not have an operator-sdk target. 
From this repo perspective, the OLM artifacts are constant and hard-coded into the code base.

nb-ohad (changes requested):

nb-ohad:
@iamniting I left a review above but I feel more context is needed. 
We made a conscious decision to go with a kubebuilder-based operator and not an operator SDK based operator. 
We do acknowledge that we need to deliver OLM artifacts like CSV and bundle to be able to publish this operator downstream, but that does not mean we want to introduce operator SDK or its toolchain into this repo. 

What I had in mind is having a folder to provide hard-coded artifacts that will serve us to create an OLM bundle and that will be similar to our approach to having a different folder to deliver this operator using helm chart and other delivery methods.

If you think that cannot be achieved in the way I describe above then please, lets have an open discussion around the subject

iamniting:
> @iamniting I left a review above but I feel more context is needed. We made a conscious decision to go with a kubebuilder-based operator and not an operator SDK based operator. We do acknowledge that we need to deliver OLM artifacts like CSV and bundle to be able to publish this operator downstream, but that does not mean we want to introduce operator SDK or its toolchain into this repo.
> 
> What I had in mind is having a folder to provide hard-coded artifacts that will serve us to create an OLM bundle and that will be similar to our approach to having a different folder to deliver this operator using helm chart and other delivery methods.
> 
> If you think that cannot be achieved in the way I describe above then please, lets have an open discussion around the subject

@nb-ohad I had a discussion with @Madhu-1 in the initial days and while discussing we decided that we would do this, It is also done in the csi-addons repo. @Madhu-1 can correct me if I am wrong. Also, I don't think we can deliver the bundle without the operator-sdk.

Can you enlighten me on how we will deliver the bundle without the operator-sdk?

nb-ohad:
Upon further inspection and research, Kustomize can be used and is part of what we already deliver with Kubebuilder. 
This means we can generate the manifest files needed, similar to how operator SDK does it. This leaves us with the operator-sdk tool. Why do we need it for the upstream project? why can we introduce it only downstream where the build system is running?

iamniting:
> Upon further inspection and research, Kustomize can be used and is part of what we already deliver with Kubebuilder. This means we can generate the manifest files needed, similar to how operator SDK does it. This leaves us with the operator-sdk tool. Why do we need it for the upstream project? why can we introduce it only downstream where the build system is running?

@nb-ohad operator-SDK is just a tool like another tool, Introducing it in the DS will make things difficult to manage in the long run. Also, we are not the only one who uses it in the upstream, other people also use it in the upstream.

Madhu-1:
> Upon further inspection and research, Kustomize can be used and is part of what we already deliver with Kubebuilder.
> This means we can generate the manifest files needed, similar to how operator SDK does it. This leaves us with the operator-sdk tool. Why do we need it for the upstream project? why can we introduce it only downstream where the build system is running?

@iamniting Yes that was the discussion before starting it and it was a very early stage, we went through a lot of iterations with the tooling. i think Rook is even having this as a downstream-only part, as we are not planning to have a bundle for the upstream. Do you still think there is a need to maintain it upstream? if yes. we can see what we can do, if there is no need we can have it downstream specific option.

iamniting:
I am closing this one. After a discussion in the mtg we came to an conclusion that the changes will be present only in the DS repo.

PR #22, opened 2024-06-25
ci: skip config folder from running codespell ci

subhamkrai:

config folder are auto generated and there is no need to
run codespell on auto generated files.

**Checklist:**

  guidelines in the [developer
* [ ] [Pending release

### DIFF
--- a/.github/workflows/lint.yaml
+++ b/.github/workflows/lint.yaml
@@ -34,7 +34,7 @@ jobs:
       - name: codespell
         uses: codespell-project/actions-codespell@master
         with:
-          skip: .git,*.sum,vendor
+          skip: .git,*.sum,vendor,config
           ignore_words_list: AfterAll
           check_filenames: true
           check_hidden: true


Madhu-1 (changes requested):
@subhamkrai please update the PR description and commit about why this change is requried.

subhamkrai:
> @subhamkrai please update the PR description and commit about why this change is requried.

done

Madhu-1:
>config folder are auto generated and there is no need to run codespell on auto generated files.

Is this true, i know that we also edit these files manually isn't it true (i think we update the sample with an example yaml)?

subhamkrai:
> > config folder are auto generated and there is no need to run codespell on auto generated files.
> 
> Is this true, i know that we also edit these files manually isn't it true (i think we update the sample with an example yaml)?

I got the knowledge from @nb-ohad

Madhu-1 (approved):

PR #26, opened 2024-06-28
Add CephCluster API (cephclusters.csi.ceph.io)

nb-ohad:

Adds new API named CephCluster (cephcluster.csi.ceph.io), capturing the setting needed to configure Ceph CSI to connect to a single Ceph Cluster 

The PR does not include CRD field validation rules.
This will be resolved in a future PR

No

## Future concerns ##

There is still an active discussion around API naming and at the time of this PR we are still open to future name changes


api/v1alpha1/cephcluster_types.go
```diff
@@ -0,0 +1,65 @@
...
+type ReadAffinitySpec struct {
+	CrushLocationLabels []string `json:"crushLocationLabels,omitempty"`
+}
+
+// CephClusterSpec defines the desired state of CephCluster
+type CephClusterSpec struct {
+	Monitors             []string              `json:"monitors"`
+	ReadAffinity         ReadAffinitySpec      `json:"readAffinity,omitempty"`
+	RbdMirrorDaemonCount int                   `json:"rbdMirrorDaemonCount,omitempty"`
+	CephConfig           map[string]CephConfig `json:"cephConfig,omitempty"`
```
Madhu-1, line 35: Let's move this to a configmap where the user will get the configmap name in the CR. WDYT?

Madhu-1 (approved):

nb-ohad:
@travisn I agree, we do need to change that sooner rather than later.
I was thinking about changing this to `CephConnection`. WDYT?

PR #32, opened 2024-07-04
ci: Adding gosec in golangci lint

subhamkrai:

ci: add gosec in golangci lint check

adding gosec in golangci lint check, also skip gosec G204 as we don't
want to run gosec on exec method

**Checklist:**

  guidelines in the [developer
* [ ] [Pending release

### DIFF
--- a/test/e2e/e2e_test.go
+++ b/test/e2e/e2e_test.go
@@ -38,6 +38,7 @@ var _ = Describe("controller", Ordered, func() {
 		Expect(utils.InstallCertManager()).To(Succeed())
 
 		By("creating manager namespace")
+		//nolint:gosec
 		cmd := exec.Command("kubectl", "create", "ns", namespace)
 		_, _ = utils.Run(cmd)
 	})
@@ -50,6 +51,7 @@ var _ = Describe("controller", Ordered, func() {
 		utils.UninstallCertManager()
 
 		By("removing manager namespace")
+		//nolint:gosec
 		cmd := exec.Command("kubectl", "delete", "ns", namespace)
 		_, _ = utils.Run(cmd)
 	})
@@ -63,6 +65,7 @@ var _ = Describe("controller", Ordered, func() {
 			var projectimage = "example.com/ceph-csi-operator:v0.0.1"
 
 			By("building the manager(Operator) image")
+			//nolint:gosec
 			cmd := exec.Command("make", "docker-build", fmt.Sprintf("IMG=%s", projectimage))
 			_, err = utils.Run(cmd)
 			ExpectWithOffset(1, err).NotTo(HaveOccurred())
@@ -72,11 +75,13 @@ var _ = Describe("controller", Ordered, func() {
 			ExpectWithOffset(1, err).NotTo(HaveOccurred())
 
 			By("installing CRDs")
+			//nolint:gosec
 			cmd = exec.Command("make", "install")
 			_, err = utils.Run(cmd)
 			ExpectWithOffset(1, err).NotTo(HaveOccurred())
 
 			By("deploying the controller-manager")
+			//nolint:gosec
 			cmd = exec.Command("make", "deploy", fmt.Sprintf("IMG=%s", projectimage))
 			_, err = utils.Run(cmd)
 			ExpectWithOffset(1, err).NotTo(HaveOccurred())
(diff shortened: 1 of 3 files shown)


nb-ohad:
@subhamkrai Can you please expend on the reason to disable G204?

subhamkrai:
> @subhamkrai Can you please expend on the reason to disable G204?

@nb-ohad gosec G204 gives error when subprocess is started with in the a process, like `exec.Command` and in other repos also upstream I have seen they suggesting to either disable gosec on that line or disable `G204`

Madhu-1:
@subhamkrai we should not disable the security rules in the whole repo, if required disable it by adding a comment where it is used or disable it in the test files (if possible)

subhamkrai:
> @subhamkrai we should not disable the security rules in the whole repo, if required disable it by adding a comment where it is used or disable it in the test files (if possible)

we can disable for single line, but this rule was something I noticed disabled most places

Madhu-1:
> > @subhamkrai we should not disable the security rules in the whole repo, if required disable it by adding a comment where it is used or disable it in the test files (if possible)
> 
> we can disable for single line, but this rule was something I noticed disabled most places

these kind of rules need to be verified and disable on demand based on the impact, For now lets not disable it by default rather disable it by putting a comment on the line in code

subhamkrai:
> > > @subhamkrai we should not disable the security rules in the whole repo, if required disable it by adding a comment where it is used or disable it in the test files (if possible)
> > 
> > 
> > we can disable for single line, but this rule was something I noticed disabled most places
> 
> these kind of rules need to be verified and disable on demand based on the impact, For now lets not disable it by default rather disable it by putting a comment on the line in code

there is small problem now, we need to put `//nolint:gosec` in the same line as `exec.command` but when `exec.command` is spilt into multiple lines is failing with G204  but if we keep `exec.command` golangci lint in same long line it will fail again due to char limit per line again golangci lint is failing

subhamkrai:
> > > > @subhamkrai we should not disable the security rules in the whole repo, if required disable it by adding a comment where it is used or disable it in the test files (if possible)
> > > 
> > > 
> > > we can disable for single line, but this rule was something I noticed disabled most places
> > 
> > 
> > these kind of rules need to be verified and disable on demand based on the impact, For now lets not disable it by default rather disable it by putting a comment on the line in code
> 
> there is small problem now, we need to put `//nolint:gosec` in the same line as `exec.command` but when `exec.command` is spilt into multiple lines is failing with G204 but if we keep `exec.command` golangci lint in same long line it will fail again due to char limit per line again golangci lint is failing

@Madhu-1 ^

nb-ohad:
@subhamkrai Where exactly do we have `exec.command` in the repo? We should not have it in code that runs on production. Do we have something like that in tests or supporting code?

test/e2e/e2e_test.go
```diff
@@ -38,7 +38,7 @@ var _ = Describe("controller", Ordered, func() {
 		Expect(utils.InstallCertManager()).To(Succeed())
 
 		By("creating manager namespace")
-		cmd := exec.Command("kubectl", "create", "ns", namespace)
+		cmd := exec.Command("kubectl", "create", "ns", namespace) //nolint:gosec
```
subhamkrai, line 41: @nb-ohad for example this place

nb-ohad: Yes, but this is test code, not code that will run as part of a csi deployment. If all of the cases are like that then we should not diable the rule globally just locally. I believe the last push did exactly that.

nb-ohad:
@subhamkrai Can you please fix the lint errors?

subhamkrai:
> @subhamkrai Can you please fix the lint errors?

@nb-ohad misspell error will go away, and for golangCI lint we need to decided how we want to fix as I mentioned 
```
there is small problem now, we need to put //nolint:gosec in the same line as exec.command but when exec.command is spilt into multiple lines is failing with G204 but if we keep exec.command golangci lint in same long line it will fail again due to char limit per line again golangci lint is failing
```
in above comment

Madhu-1:
@subhamkrai did you try to add that comment on top of that line?

test/utils/utils.go
```diff
@@ -87,12 +90,14 @@ func UninstallCertManager() {
...
 func InstallCertManager() error {
 	url := fmt.Sprintf(certmanagerURLTmpl, certmanagerVersion)
+	//nolint:gosec
 	cmd := exec.Command("kubectl", "apply", "-f", url)
 	if _, err := Run(cmd); err != nil {
 		return err
 	}
 	// Wait for cert-manager-webhook to be ready, which can take time if cert-manager
-	// was re-installed after uninstalling on a cluster.
+	// was re-installed after uninstalling on a cluster.\
```
Madhu-1, line 99: do we need `\` at the end of this line?

subhamkrai: removed

Madhu-1 (approved):

nb-ohad (approved):

PR #40, opened 2024-07-10
Add Config API (configs.csi.ceph.io)

nb-ohad:

Adds new API named Config (configs.csi.ceph.io), defining the information Ceph CSI needs to consume storage from a Ceph cluster
The PR does not include CRD field validation rules.
This will be resolved in a future PR

No


api/v1alpha1/config_types.go
```diff
@@ -0,0 +1,73 @@
...
+	RadosNamespace string `json:"radosNamespace,omitempty"`
+}
+
+// NfsConfigSpec cdefines the desired NFS configuration
+type NfsConfigSpec struct {
+}
+
+// ConfigSpec defines the desired state of Config
+type ConfigSpec struct {
+	CephClusterRef corev1.LocalObjectReference `json:"cephClusterRef,omitempty"`
```
Madhu-1, line 40: @nb-ohad can you remove omitempty from this one, we need to have it set

nb-ohad: Omit empty is a serialization directive it does not indicate that the field is optional or not, it just mean that the filed will not be serialized on write if the value is empty.
 
For an optional field, we need an `+optional` comment above the field, so OpenAPI will treat it as optional (which we don't have right now for anything).

But regardless I can remove the omitempty if you feel it is necessary

Madhu-1: okay, it looks like i was not clear, i wanted to get it removed as its deadcode as we always set this value and its never going to be empty

nb-ohad: Yep, I can do that. 
I usually set omitempty for all fields of my APIs to lower the footprint when manually marshaling (for example in unit test code)

nb-ohad: done

Madhu-1 (approved):

PR #48, opened 2024-07-21
ClientProfile reconcile

nb-ohad:
Initial implications for `clientprofiles.csi.ceph.io` controller

- Controller validation against a cephconnection.csi.ceph.io resource
- Shared owner logic for attached cephconnection.csi.ceph.io logic 
- Shared owner logic for Ceph CSI Config, config map 
- Reconcile a Ceph CSI Config record into the shared config map
- Handle clean-up upon resource deletion (removal of owners, removal of configmap record, etc.)

- PR does not include test code. Test code will be added in a new PR
- ~~Controller name might need to change as of pending discussion on `config.csi.ceph.io` name change~~


internal/controller/config_controller.go
```diff
@@ -167,6 +179,28 @@ func (r *ConfigReconcile) loadAndValidate() error {
 	return nil
 }
 
+func (r *ConfigReconcile) reconcileCephCluster() error {
+	log := r.log.WithValues("cephClusterName", r.cephCluster.Name)
+	log.Info("Reconciling Ceph Clsuter")
```
Madhu-1, line 184: `Clsuter` to `Cluster`

internal/controller/config_controller.go
```diff
@@ -106,28 +160,95 @@ func (r *ConfigReconcile) reconcile() error {
...
+		} else {
+			clusterInfoList = append(clusterInfoList, record)
+		}
+
+		// serialize the list ba
+		if bytes, err := json.Marshal(clusterInfoList); err == nil {
+			if csiConfigMap.Data == nil {
+				csiConfigMap.Data = map[string]string{}
+			}
+			csiConfigMap.Data["config.json"] = string(bytes)
```
Madhu-1, line 272: define a const for `config.json`?

internal/utils/meta.go
```diff
@@ -31,3 +33,14 @@ func AddAnnotation(obj metav1.Object, key string, value string) bool {
 	}
 	return false
 }
+
+// HasOwnerReference returns true if the object has an owner ref
```
Madhu-1, line 39: `HasOwnerReference` to `IsOwnedBy`

internal/controller/config_controller.go
```diff
@@ -1,6 +1,4 @@
 /*
-Copyright 2024.
-
```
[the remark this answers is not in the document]
nb-ohad: no, just a mistake, I will revert

internal/controller/config_controller.go
```diff
@@ -33,30 +38,96 @@ type ConfigReconciler struct {
...
-		Complete(r)
+func (r *ConfigReconcile) reconcile() error {
+	// Load the config object
+	if err := r.Get(r.ctx, client.ObjectKeyFromObject(&r.config), &r.config); err != nil {
+		r.log.Error(err, "Unable to load configs.csi.ceph.io resource")
+		return err
+	}
+
+	// Validate a pointer to a ceph cluster resource
+	if r.config.Spec.CephClusterRef.Name == "" {
```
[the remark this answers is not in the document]
nb-ohad: Yes, and I am planning on adding the proper validation to the type definition. 
With that said, there is no harm in having another check in-code

internal/utils/meta.go
nb-ohad: Good catch
I started with the `HasOwnerReference` name but it is too generic as it means "has any owner ref", then I changed it to `IsOwnedBy`. It seems I forgot to update the comment

internal/controller/config_controller.go
nb-ohad: I didn't see a reason, but you mentioned it so I will add a const.

internal/controller/config_controller.go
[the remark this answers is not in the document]
nb-ohad: done

internal/controller/config_controller.go
nb-ohad: done

internal/controller/config_controller.go
nb-ohad: done

Madhu-1:
@nb-ohad can you please resolve merge conflicts?

leelavg:
can below be included in this PR?
``` diff
diff --git a/config/rbac/kustomization.yaml b/config/rbac/kustomization.yaml
index c26ae59..3d0a21b 100644
--- a/config/rbac/kustomization.yaml
+++ b/config/rbac/kustomization.yaml
@@ -28,3 +28,21 @@ resources:
 - operatorconfig_viewer_role.yaml
 - driver_editor_role.yaml
 - driver_viewer_role.yaml
+- csi_cephfs_ctrlplugin_clusterrole.yaml
+- csi_cephfs_ctrlplugin_clusterrole_binding.yaml
+- csi_cephfs_ctrlplugin_role.yaml
+- csi_cephfs_ctrlplugin_role_binding.yaml
+- csi_cephfs_ctrlplugin_service_account.yaml
+- csi_cephfs_nodeplugin_clusterrole.yaml
+- csi_cephfs_nodeplugin_clusterrole_binding.yaml
+- csi_cephfs_nodeplugin_service_account.yaml
+- csi_rbd_ctrlplugin_clusterrole.yaml
+- csi_rbd_ctrlplugin_clusterrole_binding.yaml
+- csi_rbd_ctrlplugin_role.yaml
+- csi_rbd_ctrlplugin_role_binding.yaml
+- csi_rbd_ctrlplugin_service_account.yaml
+- csi_rbd_nodeplugin_clusterrole.yaml
+- csi_rbd_nodeplugin_clusterrole_binding.yaml
+- csi_rbd_nodeplugin_role.yaml
+- csi_rbd_nodeplugin_role_binding.yaml
+- csi_rbd_nodeplugin_service_account.yaml
```

internal/controller/config_controller.go
```diff
@@ -65,6 +91,11 @@ func (r *ConfigReconciler) SetupWithManager(mgr ctrl.Manager) error {
 			builder.MatchEveryOwner,
 			builder.WithPredicates(genChangedPredicate),
 		).
+		Owns(
+			&corev1.ConfigMap{},
```
leelavg, line 100: missing rbac for this.

internal/controller/config_controller.go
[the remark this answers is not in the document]
nb-ohad: Validation was added here: https://github.com/ceph/ceph-csi-operator/pull/62

internal/controller/config_controller.go
nb-ohad: Done

api/v1alpha1/cephcluster_types.go
```diff
@@ -27,9 +27,9 @@ type ReadAffinitySpec struct {
 
 // CephClusterSpec defines the desired state of CephCluster
 type CephClusterSpec struct {
-	Monitors             []string         `json:"monitors"`
-	ReadAffinity         ReadAffinitySpec `json:"readAffinity,omitempty"`
-	RbdMirrorDaemonCount int              `json:"rbdMirrorDaemonCount,omitempty"`
+	Monitors             []string          `json:"monitors"`
```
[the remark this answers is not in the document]
nb-ohad: done

nb-ohad:
@Madhu-1 I am adding the hold label on this one to make sure that https://github.com/ceph/ceph-csi-operator/pull/65 is merged first. That PR contains a small fix for map-to-string serialization needed in the code supporting Kernal and Fuse mount options.

I cannot just rebase this PR on top of the fix PR because this PR is already rebased on top of the API changes which is not yet merged

nb-ohad:
> can below be included in this PR?
> 
> ```diff
> diff --git a/config/rbac/kustomization.yaml b/config/rbac/kustomization.yaml
> index c26ae59..3d0a21b 100644
> --- a/config/rbac/kustomization.yaml
> +++ b/config/rbac/kustomization.yaml
> @@ -28,3 +28,21 @@ resources:
>  - operatorconfig_viewer_role.yaml
>  - driver_editor_role.yaml
>  - driver_viewer_role.yaml
> +- csi_cephfs_ctrlplugin_clusterrole.yaml
> +- csi_cephfs_ctrlplugin_clusterrole_binding.yaml
> +- csi_cephfs_ctrlplugin_role.yaml
> +- csi_cephfs_ctrlplugin_role_binding.yaml
> +- csi_cephfs_ctrlplugin_service_account.yaml
> +- csi_cephfs_nodeplugin_clusterrole.yaml
> +- csi_cephfs_nodeplugin_clusterrole_binding.yaml
> +- csi_cephfs_nodeplugin_service_account.yaml
> +- csi_rbd_ctrlplugin_clusterrole.yaml
> +- csi_rbd_ctrlplugin_clusterrole_binding.yaml
> +- csi_rbd_ctrlplugin_role.yaml
> +- csi_rbd_ctrlplugin_role_binding.yaml
> +- csi_rbd_ctrlplugin_service_account.yaml
> +- csi_rbd_nodeplugin_clusterrole.yaml
> +- csi_rbd_nodeplugin_clusterrole_binding.yaml
> +- csi_rbd_nodeplugin_role.yaml
> +- csi_rbd_nodeplugin_role_binding.yaml
> +- csi_rbd_nodeplugin_service_account.yaml
> ```

@leelavg is this request still relevant? Aren't these changes already in main?

leelavg:
>  is this request still relevant? Aren't these changes already in main?
- Yep, should have been part of fixes PR that was merged. In main the manifests are defined but will not be included in the bundle unless referenced in Kustomization.yml

nb-ohad:
> > is this request still relevant? Aren't these changes already in main?
> 
> * Yep, should have been part of fixes PR that was merged. In main the manifests are defined but will not be included in the bundle unless referenced in Kustomization.yml

This request is out of scope of this PR. If you have the solution please provide a PR. If not please open a gitissue to track

nb-ohad:
@Madhu-1 
PR refreshed, hold removed.

internal/controller/config_controller.go
leelavg: resolve.

internal/controller/clientprofile_controller.go
```diff
@@ -0,0 +1,332 @@
...
+	record.ClusterId = clientProfile.Name
+	record.Monitors = cephConn.Spec.Monitors
+	if cephFs := clientProfile.Spec.CephFs; cephFs != nil {
+		record.CephFs.SubvolumeGroup = cephFs.SubVolumeGroup
+		if mountOpt := cephFs.KernelMountOptions; mountOpt != nil {
+			record.CephFs.KernelMountOptions = utils.MapToString(mountOpt, "=", ",")
+		}
+		if mountOpt := cephFs.FuseMountOptions; mountOpt != nil {
+			record.CephFs.FuseMountOptions = utils.MapToString(mountOpt, "=", ",")
+		}
```
Madhu-1, line 321: i forgot the reason to make the KernelMountOptions and FuseMountOptions as map why not string?

nb-ohad: Some mount options are only keys, some are key-value pairs, and the operator is always fixed.

We wanted to minimize the user's chance of error in formatting the string. By providing a map interface the user does not need to concern themselves with serialization concerns.

internal/controller/clientprofile_controller.go
```diff
@@ -0,0 +1,332 @@
...
+	}
+
+	return nil
+}
+
+func (r *ClientProfileReconcile) loadAndValidate() error {
+	// Load the config object
+	if err := r.Get(r.ctx, client.ObjectKeyFromObject(&r.clientProfile), &r.clientProfile); err != nil {
+		r.log.Error(err, "Failed loading configs.csi.ceph.io resource")
```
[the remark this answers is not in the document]
nb-ohad: I moved away from FQNs in error messages. I will change to `ClientProfile`

internal/controller/clientprofile_controller.go
Madhu-1: If we allow this we might get into a problem if the format we are passing to the ceph kernel or fuse after the operator is formatting the user input, IMO we don't need to over-simplify this for the user and the admin knows what exactly needs to be set and the operator should just set that in the configmap. my suggestion would be to make it string and pass it to the configmap.

internal/controller/clientprofile_controller.go
```diff
@@ -0,0 +1,332 @@
...
+	CephFs    struct {
+		SubvolumeGroup     string `json:"subvolumeGroup,omitempty"`
+		KernelMountOptions string `json:"kernelMountOptions"`
+		FuseMountOptions   string `json:"fuseMountOptions"`
+	} `json:"cephFS,omitempty"`
+	Rbd struct {
+		RadosNamespace string `json:"radosNamespace,omitempty"`
+		MirrorCount    int    `json:"mirrorCount,omitempty"`
+	} `json:"rbd,omitempty"`
+	Nfs          struct{} `json:"nfs,omitempty"`
```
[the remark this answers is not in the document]
nb-ohad: I don't understand what you mean.  
Do you mean that NFS lacks support in Ceph CSI's cluster record? 

I can see support for in ceph csi code: https://github.com/ceph/ceph-csi/blob/devel/api/deploy/kubernetes/csi-config-map.go#L19-L32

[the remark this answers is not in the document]
Madhu-1: It's an extra which will be present but we will not use it. Resolving let's keep it and we will remove it later once this is removed from cephcsi

internal/controller/clientprofile_controller.go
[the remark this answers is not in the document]
nb-ohad: done

internal/controller/clientprofile_controller.go
```diff
@@ -0,0 +1,332 @@
...
+	// Load the ceph connection
+	r.cephConn.Name = r.clientProfile.Spec.CephConnectionRef.Name
+	r.cephConn.Namespace = r.clientProfile.Namespace
+	if err := r.Get(r.ctx, client.ObjectKeyFromObject(&r.cephConn), &r.cephConn); err != nil {
+		r.log.Error(err, "Failed loading CephConnection")
+		return err
+	}
+
+	// Ensure the CephConnection has an owner reference (not controller reference)
+	// for the current reconciled config resource
```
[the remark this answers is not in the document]
nb-ohad: done

internal/controller/clientprofile_controller.go
```diff
@@ -0,0 +1,332 @@
...
+	defer configMapUpdateLock.Unlock()
+
+	_, err := ctrlutil.CreateOrUpdate(r.ctx, r.Client, &csiConfigMap, func() error {
+		if _, err := utils.ToggleOwnerReference(
+			!r.cleanUp,
+			&r.clientProfile,
+			&csiConfigMap,
+			r.Scheme,
+		); err != nil {
+			log.Error(err, "Failed togggling owner reference for Ceph CSI config map")
```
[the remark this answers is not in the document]
nb-ohad: done

internal/controller/clientprofile_controller.go
```diff
@@ -0,0 +1,337 @@
...
+		Enabled             bool     `json:"enabled,omitempty"`
+		CrushLocationLabels []string `json:"crushLocationLabels,omitempty"`
+	} `json:"readAffinity,omitempty"`
+}
+
+const (
+	cleanupFinalizer = "csi.ceph.com/cleanup"
+)
+
+var configMapUpdateLock = sync.Mutex{}
```
leelavg, line 87: iirc, zero value of sync.Mutex also works.

internal/controller/clientprofile_controller.go
```diff
@@ -0,0 +1,337 @@
...
+			return err
+		}
+	}
+
+	return nil
+}
+
+func (r *ClientProfileReconcile) loadAndValidate() error {
+	// Load the ClientProfile
+	if err := r.Get(r.ctx, client.ObjectKeyFromObject(&r.clientProfile), &r.clientProfile); err != nil {
```
leelavg, line 169: I see a pattern of loading operands across all controllers, however an event is always being assumed as the resource will exist but for deletion it may not and this could potentially result in not ending reconciles, isn't it?

internal/controller/clientprofile_controller.go
```diff
@@ -0,0 +1,337 @@
...
+		r.log.Error(err, "Failed loading ClientProfile")
+		return err
+	}
+	r.cleanUp = r.clientProfile.DeletionTimestamp != nil
+
+	// Validate a pointer to a ceph cluster resource
+	if r.clientProfile.Spec.CephConnectionRef.Name == "" {
```
leelavg, line 176: any reason in not using a CEL rule?

internal/controller/clientprofile_controller.go
```diff
@@ -0,0 +1,337 @@
...
+	// Using a lock to serialized the updating of the config map.
+	// Although the code will run perfetcly fine without the lock, there will be a higher
+	// chance to fail on the create/update operation because another concurrent reconcile loop
+	// updated the config map which will result in stale representation and an update failure.
+	// The locking strategy will sync all update to the shared config file and will prevent such
+	// potential issues without a big impact on preformace as a whole
+	configMapUpdateLock.Lock()
+	defer configMapUpdateLock.Unlock()
+
+	_, err := ctrlutil.CreateOrUpdate(r.ctx, r.Client, &csiConfigMap, func() error {
```
leelavg, line 254: so the configmap creation happen only after profile is created, were any changes already made that driver/operatorconfig doesn't look for this before they deploy CSI?

internal/utils/meta.go
```diff
@@ -31,3 +35,35 @@ func AddAnnotation(obj metav1.Object, key string, value string) bool {
...
+	return false
+}
+
+// ToggleOwnerReference adds or remove an owner reference for the given owner based on the first argument.
+// The function return true if the owner reference list had changed and false it it didn't
+func ToggleOwnerReference(on bool, obj, owner metav1.Object, scheme *runtime.Scheme) (bool, error) {
+	ownerRefExists := IsOwnedBy(obj, owner)
+
+	if on {
+		if !ownerRefExists {
```
leelavg, line 56: any reason in not clubbing this condition?

internal/utils/meta.go
```diff
@@ -31,3 +35,35 @@ func AddAnnotation(obj metav1.Object, key string, value string) bool {
...
+		if owner.GetUID() == ownerRefs[i].UID {
+			return true
+		}
+	}
+	return false
+}
+
+// ToggleOwnerReference adds or remove an owner reference for the given owner based on the first argument.
+// The function return true if the owner reference list had changed and false it it didn't
+func ToggleOwnerReference(on bool, obj, owner metav1.Object, scheme *runtime.Scheme) (bool, error) {
```
leelavg, line 52: is my understanding is correct? we are manipulating owners as not to remove the dependent if last owner is deleted?

internal/controller/clientprofile_controller.go
nb-ohad: Why do you prefer 
```
```suggestion
var configMapUpdateLock sync.Mutex
```
To
```suggestion
var configMapUpdateLock = sync.Mutex{}
```

internal/controller/clientprofile_controller.go
nb-ohad: A CEL rule is not a replacement for double-checking in code. 
Are we missing a CEL rule?

internal/controller/clientprofile_controller.go
nb-ohad: This is something the was already discussed and I will provide these changes in a future PR

internal/utils/meta.go
nb-ohad: If not `on` we should execute `} else if ownerRefExists {` (line 62)
If `on` but  `!ownerRefExists` we should not execute `} else if ownerRefExists {` (line 62)

internal/utils/meta.go
nb-ohad: This is just a util that adds or removes from the owner list based on a provided flag

internal/controller/clientprofile_controller.go
leelavg: Not using something that isn't required.

internal/controller/clientprofile_controller.go
leelavg: I think we are missing CEL. xxRef struct is made required but the field Name in LocalRef is optional for backwards compat and so CEL rule on our struct would be good.

internal/controller/clientprofile_controller.go
leelavg: Ack, resolve.

internal/utils/meta.go
leelavg: Yes, I see atm this is being used for manipulating owners, so want to understand why do we need to explictly remove owner.

internal/utils/meta.go
leelavg: Yes, `if on && !ownerRefExists` would result in the same because `on` acting as short-circuit, isn't it?

If we want to be elaborated it's fine, just when I saw, it occured to me that I might be missing something.

internal/utils/meta.go
nb-ohad: For the cases when someone deleted the ClientProfile. We don't want to delete all owners and we don't want a "zombie" ownership that will prevent proper cleanup

internal/utils/meta.go
nb-ohad: @leelavg 
```
if on {
    if !ownerRefExists {
    }
} else ownerRefExists {
}
```
is not the same as 
```

if on && !ownerRefExists {
} else ownerRefExists {
}
```

Consider the case where on `on = true` and `ownerRefExists = true`. 
In the first snippet (my code) we will skip everything, with the second snippet (your proposal) we will run the code inside the else (because `ownerRefExists = true`)

internal/controller/clientprofile_controller.go
nb-ohad: I am not aware the CEL can look into structs that we do not define (in this case it is `LocalObjectReference` object)
If there is a way to achieve that, please point me in the write direction

internal/controller/clientprofile_controller.go
nb-ohad: I am willing to change, but this is mainly styling. The generated code will be exactly the same

internal/controller/clientprofile_controller.go
```diff
@@ -0,0 +1,337 @@
...
+	"sigs.k8s.io/controller-runtime/pkg/predicate"
+
+	csiv1a1 "github.com/ceph/ceph-csi-operator/api/v1alpha1"
+	"github.com/ceph/ceph-csi-operator/internal/utils"
+)
+
+//+kubebuilder:rbac:groups=csi.ceph.io,resources=clientprofile,verbs=get;list;watch;create;update;patch;delete
+//+kubebuilder:rbac:groups=csi.ceph.io,resources=clientprofile/status,verbs=get;update;patch
+//+kubebuilder:rbac:groups=csi.ceph.io,resources=clientprofile/finalizers,verbs=update
+//+kubebuilder:rbac:groups=csi.ceph.io,resources=cephconnection,verbs=get;list;watch;update
```
leelavg, line 43: resources should be plural. clientprofile to clientprofiles and cephconnection to cephconnections.

internal/controller/clientprofile_controller.go
```diff
@@ -0,0 +1,337 @@
...
+	cleanupFinalizer = "csi.ceph.com/cleanup"
+)
+
+var configMapUpdateLock = sync.Mutex{}
+
+//+kubebuilder:rbac:groups=csi.ceph.io,resources=clientprofile,verbs=get;list;watch;create;update;patch;delete
+//+kubebuilder:rbac:groups=csi.ceph.io,resources=clientprofile/status,verbs=get;update;patch
+//+kubebuilder:rbac:groups=csi.ceph.io,resources=clientprofile/finalizers,verbs=update
+//+kubebuilder:rbac:groups=csi.ceph.io,resources=cephconnection,verbs=get;list;watch;update
+//+kubebuilder:rbac:groups="",resources=configmap,verbs=get;list;watch;create;update;patch;delete
```
leelavg, line 93: resources should be plural. clientprofile to clientprofiles and cephconnection to cephconnections, configmap to configmaps

internal/utils/meta.go
leelavg: > We don't want to delete all owners
- there is no need to delete the owner, isn't it?

> we don't want a "zombie" ownership that will prevent proper cleanup
- when all the owners are deleted then only the dependent is removed by k8s and the lib (ctrlutil) afaik takes care of updating the owner w/ same name but different uid

internal/utils/meta.go
leelavg: > we will skip everything
- may not skip everything, however I understood the explanation, resolve.

internal/controller/clientprofile_controller.go
nb-ohad: We agreed to shift to an array approach with a formating specifier. I will add it in a future PR.
@Madhu-1 Can we resolve this?

internal/controller/clientprofile_controller.go
Madhu-1: As discussed https://github.com/ceph/ceph-csi-operator/issues/78 opened a tracker for it. resolving it

internal/controller/clientprofile_controller.go
nb-ohad: Fixed now

internal/controller/clientprofile_controller.go
nb-ohad: Fixed now

Madhu-1 (approved):

internal/controller/clientprofile_controller.go
leelavg: resolve.

PR #50, opened 2024-07-22
logrotate: add logrotate functionality for csi

parth-gr:

1) Make main container and csi addons container
   log to a file(dependency on klog)

2) Add a log-rotate sidecar container,
   so it can rotate the logs

3) Added other volume and volumemounts as needed

4) Added the privileged option for controllerplugin

Do you have any questions?

For example:

## Related issues ##

Fixes: #issue_number

## Future concerns ##

**Checklist:**

  guidelines in the [developer
* [ ] [Pending release

### DIFF
--- a/api/v1alpha1/driver_types.go
+++ b/api/v1alpha1/driver_types.go
@@ -33,8 +33,10 @@ const (
 	MonthlyPeriod PeriodicityType = "monthly"
 )
 
+// +kubebuilder:validation:XValidation:message="Either maxLogSize or periodicity must be set",rule="(has(self.maxLogSize)) || (has(self.periodicity))"
 type LogRotationSpec struct {
 	// MaxFiles is the number of logrtoate files
+	// Default to 7
 	//+kubebuilder:validation:Optional
 	MaxFiles int `json:"maxFiles,omitempty"`
 
@@ -48,6 +50,7 @@ type LogRotationSpec struct {
 	Periodicity PeriodicityType `json:"periodicity,omitempty"`
 
 	// LogHostPath is the prefix directory path for the csi log files
+	// Default to /var/lib/cephcsi
 	//+kubebuilder:validation:Optional
 	LogHostPath string `json:"logHostPath,omitempty"`
 }
--- a/internal/controller/driver_controller.go
+++ b/internal/controller/driver_controller.go
@@ -230,6 +230,7 @@ func (r *driverReconcile) reconcile() error {
 	// the desired state defined on the driver object
 	errChan := utils.RunConcurrently(
 		r.reconcileCsiConfigMap,
+		r.reconcileLogRotateConfigMap,
 		r.reconcileK8sCsiDriver,
 		r.reconcileControllerPluginDeployment,
 		r.reconcileNodePluginDeamonSet,
@@ -325,6 +326,63 @@ func (r *driverReconcile) LoadAndValidateDesiredState() error {
 	}
 
 	return nil
+}
+
+func (r *driverReconcile) reconcileLogRotateConfigMap() error {
+	logRotateConfigmap := &corev1.ConfigMap{}
+	logRotateConfigmap.Name = utils.LogRotateConfigMapName(r.driver.Name)
+	logRotateConfigmap.Namespace = r.driver.Namespace
+
+	log := r.log.WithValues("logRotateConfigMap", logRotateConfigmap.Name)
+	log.Info("Reconciling logRotate configmap")
+
+	logRotationSpec := cmp.Or(r.driver.Spec.Log, &csiv1a1.LogSpec{}).Rotation
+	if logRotationSpec != nil {
+		opResult, err := ctrlutil.CreateOrUpdate(r.ctx, r.Client, logRotateConfigmap, func() error {
+			if _, err := utils.ToggleOwnerReference(true, logRotateConfigmap, &r.driver, r.Scheme); err != nil {
+				log.Error(err, "Failed adding an owner reference on the LogRotate config map")
+				return err
+			}
+			if logRotationSpec.Periodicity != "" && logRotationSpec.MaxLogSize.IsZero() {
+				err := fmt.Errorf("invalid Log.Rotation spec")
+				log.Error(err, "Either \"maxLogSize\" or \"periodicity\" fields must be set")
+				return err
+			}
+			maxFiles := cmp.Or(logRotationSpec.MaxFiles, defaultLogRotateMaxFiles)
+			cronLogRotateSettings := []string{
+				"\tmissingok",
+				"\tcompress",
+				"\tcopytruncate",
+				"\tnotifempty",
+				fmt.Sprintf("\trotate %d", maxFiles),
+			}
+			if logRotationSpec.Periodicity != "" {
+				periodicity := "\t" + string(logRotationSpec.Periodicity)
+				cronLogRotateSettings = append(cronLogRotateSettings, periodicity)
+			}
+			if logRotationSpec.MaxLogSize.String() != "0" {
+				maxSize := fmt.Sprintf("\tmaxsize %s", logRotationSpec.MaxLogSize.String())
+				cronLogRotateSettings = append(cronLogRotateSettings, maxSize)
+			}
+			logRotateConfigmap.Data = map[string]string{
+				"csi": fmt.Sprintf(
+					"/csi-logs/*.log {\n%s\n}\n",
+					strings.Join(cronLogRotateSettings, "\n"),
+				),
+			}
+			return nil
+		})
+
+		logCreateOrUpdateResult(log, "LogRotateConfigMap", logRotateConfigmap, opResult, err)
+		return err
+	} else {
+		// Remove the logrotate configmap if logrotate setting is removed from the driver's spec
+		if err := r.Delete(r.ctx, logRotateConfigmap); client.IgnoreNotFound(err) != nil {
+			log.Error(err, "Unable to delete LogRotate configmap")
+			return err
+		}
+		return nil
+	}
 }
 
 func (r *driverReconcile) reconcileCsiConfigMap() error {
@@ -454,6 +512,10 @@ func (r *driverReconcile) reconcileControllerPluginDeployment() error {
 		logVerbosity := ptr.Deref(r.driver.Spec.Log, csiv1a1.LogSpec{}).Verbosity
 		forceKernelClient := r.isCephFsDriver() && r.driver.Spec.CephFsClientType == csiv1a1.KernelCephFsClient
 		snPolicy := cmp.Or(r.driver.Spec.SnapshotPolicy, csiv1a1.VolumeSnapshotSnapshotPolicy)
+		logRotationSpec := cmp.Or(r.driver.Spec.Log, &csiv1a1.LogSpec{}).Rotation
+		logHostPath := cmp.Or(logRotationSpec.LogHostPath, defaultLogHostPath)
+		logRotationEnabled := logRotationSpec != nil
+		securityContext := utils.If(logRotationEnabled, &corev1.SecurityContext{Privileged: ptr.To(true)}, nil)
 
 		leaderElectionSettingsArg := []string{
 			utils.LeaderElectionNamespaceContainerArg(r.driver.Namespace),
@@ -487,6 +549,7 @@ func (r *driverReconcile) reconcileControllerPluginDeployment() error {
 								Name:            fmt.Sprintf("csi-%splugin", r.driverType),
 								Image:           r.images["plugin"],
 								ImagePullPolicy: imagePullPolicy,
+								SecurityContext: securityContext,
 								Args: utils.DeleteZeroValues(
 									[]string{
 										utils.TypeContainerArg(string(r.driverType)),
@@ -502,6 +565,13 @@ func (r *driverReconcile) reconcileControllerPluginDeployment() error {
 										utils.If(
 											ptr.Deref(r.driver.Spec.DeployCsiAddons, false),
 											utils.CsiAddonsEndpointContainerArg,
+											"",
+										),
+										utils.If(logRotationEnabled, utils.LogToStdErrContainerArg, ""),
+										utils.If(logRotationEnabled, utils.AlsoLogToStdErrContainerArg, ""),
+										utils.If(
+											logRotationEnabled,
+											utils.LogFileContainerArg(fmt.Sprintf("csi-%splugin", r.driverType)),
 											"",
 										),
 									},
@@ -533,6 +603,9 @@ func (r *driverReconcile) reconcileControllerPluginDeployment() error {
 									}
 									if r.isRdbDriver() {
 										mounts = append(mounts, utils.OidcTokenVolumeMount)
+									}
+									if logRotationEnabled {
+										mounts = append(mounts, utils.LogsDirVolumeMount)
 									}
 									return mounts
 								}),
@@ -654,6 +727,7 @@ func (r *driverReconcile) reconcileControllerPluginDeployment() error {
 								Name:            "csi-addons",
 								Image:           r.images["addons"],
 								ImagePullPolicy: imagePullPolicy,
+								SecurityContext: securityContext,
 								Args: utils.DeleteZeroValues(
 									append(
 										slices.Clone(leaderElectionSettingsArg),
@@ -664,6 +738,9 @@ func (r *driverReconcile) reconcileControllerPluginDeployment() error {
 										utils.CsiAddonsAddressContainerArg,
 										utils.ControllerPortContainerArg,
 										utils.NamespaceContainerArg,
+										utils.If(logRotationEnabled, utils.LogToStdErrContainerArg, ""),
+										utils.If(logRotationEnabled, utils.AlsoLogToStdErrContainerArg, ""),
+										utils.If(logRotationEnabled, utils.LogFileContainerArg("csi-addons"), ""),
 									),
 								),
 								Ports: []corev1.ContainerPort{
@@ -675,9 +752,15 @@ func (r *driverReconcile) reconcileControllerPluginDeployment() error {
 									utils.PodNameEnvVar,
 									utils.PodNamespaceEnvVar,
 								},
-								VolumeMounts: []corev1.VolumeMount{
-									utils.SocketDirVolumeMount,
-								},
+								VolumeMounts: utils.Call(func() []corev1.VolumeMount {
+									mounts := []corev1.VolumeMount{
+										utils.SocketDirVolumeMount,
+									}
+									if logRotationEnabled {
+										mounts = append(mounts, utils.LogsDirVolumeMount)
+									}
+									return mounts
+								}),
 								Resources: ptr.Deref(
 									pluginSpec.Resources.Addons,
 									corev1.ResourceRequirements{},
@@ -741,7 +824,22 @@ func (r *driverReconcile) reconcileControllerPluginDeployment() error {
 								),
 							})
 						}
-
+						// CSI LogRotate Container
+						if logRotationEnabled {
+							resources := ptr.Deref(pluginSpec.Resources.LogRotator, corev1.ResourceRequirements{})
+							containers = append(containers, corev1.Container{
+								Name:            "log-rotator",
+								Image:           r.images["plugin"],
+								ImagePullPolicy: imagePullPolicy,
+								Resources:       resources,
+								Command:         []string{"/bin/bash", "-c"},
+								Args:            []string{`while true; do logrotate --verbose /logrotate-config/csi; sleep 15m; done`},
+								VolumeMounts: []corev1.VolumeMount{
+									utils.LogsDirVolumeMount,
+									utils.LogRotateDirVolumeMount,
+								},
+							})
+						}
 						return containers
 					}),
 					Volumes: utils.Call(func() []corev1.Volume {
@@ -766,6 +864,13 @@ func (r *driverReconcile) reconcileControllerPluginDeployment() error {
 							volumes = append(
 								volumes,
 								utils.KmsConfigVolume(&r.driver.Spec.Encryption.ConfigMapRef))
+						}
+						if logRotationEnabled {
+							volumes = append(
+								volumes,
+								utils.LogsDirVolume(logHostPath, deploy.Name),
+								utils.LogRotateDirVolumeName(r.driver.Name),
+							)
 						}
 						return volumes
 					}),
@@ -807,6 +912,9 @@ func (r *driverReconcile) reconcileNodePluginDeamonSet() error {
 
 		topology := r.isRdbDriver() && pluginSpec.Topology != nil
 		domainLabels := cmp.Or(pluginSpec.Topology, &csiv1a1.TopologySpec{}).DomainLabels
+		logRotationSpec := cmp.Or(r.driver.Spec.Log, &csiv1a1.LogSpec{}).Rotation
+		logHostPath := cmp.Or(logRotationSpec.LogHostPath, defaultLogHostPath)
+		logRotationEnabled := logRotationSpec != nil
 
 		daemonSet.Spec = appsv1.DaemonSetSpec{
 			Selector: &metav1.LabelSelector{
@@ -884,6 +992,13 @@ func (r *driverReconcile) reconcileNodePluginDeamonSet() error {
 											utils.DomainLabelsContainerArg(domainLabels),
 											"",
 										),
+										utils.If(logRotationEnabled, utils.LogToStdErrContainerArg, ""),
+										utils.If(logRotationEnabled, utils.AlsoLogToStdErrContainerArg, ""),
+										utils.If(
+											logRotationEnabled,
+											utils.LogFileContainerArg(fmt.Sprintf("csi-%splugin", r.driverType)),
+											"",
+										),
 									},
 								),
 								Env: []corev1.EnvVar{
@@ -911,6 +1026,9 @@ func (r *driverReconcile) reconcileNodePluginDeamonSet() error {
 									}
 									if r.isRdbDriver() {
 										mounts = append(mounts, utils.OidcTokenVolumeMount)
+									}
+									if logRotationEnabled {
+										mounts = append(mounts, utils.LogsDirVolumeMount)
 									}
 									return mounts
 								}),
@@ -965,6 +1083,7 @@ func (r *driverReconcile) reconcileNodePluginDeamonSet() error {
 								Args: utils.DeleteZeroValues(
 									[]string{
 										utils.CsiAddonsNodeIdContainerArg,
+										utils.NodeIdContainerArg,
 										utils.LogVerbosityContainerArg(logVerbosity),
 										utils.CsiAddonsAddressContainerArg,
 										utils.ControllerPortContainerArg,
@@ -972,6 +1091,9 @@ func (r *driverReconcile) reconcileNodePluginDeamonSet() error {
 										utils.NamespaceContainerArg,
 										utils.PodUidContainerArg,
 										utils.StagingPathContainerArg(kubeletDirPath),
+										utils.If(logRotationEnabled, utils.LogToStdErrContainerArg, ""),
+										utils.If(logRotationEnabled, utils.AlsoLogToStdErrContainerArg, ""),
+										utils.If(logRotationEnabled, utils.LogFileContainerArg("csi-addons"), ""),
 									},
 								),
 								Ports: []corev1.ContainerPort{
@@ -983,9 +1105,15 @@ func (r *driverReconcile) reconcileNodePluginDeamonSet() error {
 									utils.PodNamespaceEnvVar,
 									utils.PodUidEnvVar,
 								},
-								VolumeMounts: []corev1.VolumeMount{
-									utils.PluginDirVolumeMount,
-								},
+								VolumeMounts: utils.Call(func() []corev1.VolumeMount {
+									mounts := []corev1.VolumeMount{
+										utils.PluginDirVolumeMount,
+									}
+									if logRotationEnabled {
+										mounts = append(mounts, utils.LogsDirVolumeMount)
+									}
+									return mounts
+								}),
 								Resources: ptr.Deref(
 									pluginSpec.Resources.Addons,
 									corev1.ResourceRequirements{},
@@ -1026,6 +1154,22 @@ func (r *driverReconcile) reconcileNodePluginDeamonSet() error {
 								})
 							}
 						}
+						// CSI LogRotate Container
+						if logRotationEnabled {
+							resources := ptr.Deref(pluginSpec.Resources.LogRotator, corev1.ResourceRequirements{})
+							containers = append(containers, corev1.Container{
+								Name:            "log-rotator",
+								Image:           r.images["plugin"],
+								ImagePullPolicy: imagePullPolicy,
+								Resources:       resources,
+								Command:         []string{"/bin/bash", "-c"},
+								Args:            []string{`while true; do logrotate --verbose /logrotate-config/csi; sleep 15m; done`},
+								VolumeMounts: []corev1.VolumeMount{
+									utils.LogsDirVolumeMount,
+									utils.LogRotateDirVolumeMount,
+								},
+							})
+						}
 						return containers
 					}),
 					Volumes: utils.Call(func() []corev1.Volume {
@@ -1065,6 +1209,13 @@ func (r *driverReconcile) reconcileNodePluginDeamonSet() error {
 							volumes = append(
 								volumes,
 								utils.OidcTokenVolume,
+							)
+						}
+						if logRotationEnabled {
+							volumes = append(
+								volumes,
+								utils.LogsDirVolume(logHostPath, daemonSet.Name),
+								utils.LogRotateDirVolumeName(r.driver.Name),
 							)
 						}
 						return volumes
@@ -1273,6 +1424,9 @@ func mergeDriverSpecs(dest, src *csiv1a1.DriverSpec) {
 			if dest.Resources.Plugin == nil {
 				dest.Resources.Plugin = src.Resources.Plugin
 			}
+			if dest.Resources.LogRotator == nil {
+				dest.Resources.LogRotator = src.Resources.LogRotator
+			}
 		}
 	}
 	if src.ControllerPlugin != nil {
@@ -1327,6 +1481,9 @@ func mergeDriverSpecs(dest, src *csiv1a1.DriverSpec) {
 			}
 			if dest.Resources.Plugin == nil {
 				dest.Resources.Plugin = src.Resources.Plugin
+			}
+			if dest.Resources.LogRotator == nil {
+				dest.Resources.LogRotator = src.Resources.LogRotator
 			}
 		}
 	}
--- a/internal/controller/driver_controller.go
+++ b/internal/controller/driver_controller.go
@@ -69,8 +69,11 @@ const (
 	NfsDriverType    = "nfs"
 )
 
-// Annotation name for ownerref information
-const ownerRefAnnotationKey = "csi.ceph.io/ownerref"
+const (
+	// Annotation name for ownerref information
+	ownerRefAnnotationKey = "csi.ceph.io/ownerref"
+	logRotateCmd          = `while true; do logrotate --verbose /logrotate-config/csi; sleep 15m; done`
+)
 
 // A regexp used to parse driver's prefix and type from the full name
 var nameRegExp, _ = regexp.Compile(fmt.Sprintf(
@@ -513,9 +516,17 @@ func (r *driverReconcile) reconcileControllerPluginDeployment() error {
 		forceKernelClient := r.isCephFsDriver() && r.driver.Spec.CephFsClientType == csiv1a1.KernelCephFsClient
 		snPolicy := cmp.Or(r.driver.Spec.SnapshotPolicy, csiv1a1.VolumeSnapshotSnapshotPolicy)
 		logRotationSpec := cmp.Or(r.driver.Spec.Log, &csiv1a1.LogSpec{}).Rotation
-		logHostPath := cmp.Or(logRotationSpec.LogHostPath, defaultLogHostPath)
 		logRotationEnabled := logRotationSpec != nil
-		securityContext := utils.If(logRotationEnabled, &corev1.SecurityContext{Privileged: ptr.To(true)}, nil)
+		logRotateSecurityContext := utils.If(
+			pluginSpec.Privileged != nil && logRotationEnabled,
+			&corev1.SecurityContext{
+				Privileged: pluginSpec.Privileged,
+				Capabilities: &corev1.Capabilities{
+					Drop: []corev1.Capability{"All"},
+				},
+			},
+			nil,
+		)
 
 		leaderElectionSettingsArg := []string{
 			utils.LeaderElectionNamespaceContainerArg(r.driver.Namespace),
@@ -549,7 +560,7 @@ func (r *driverReconcile) reconcileControllerPluginDeployment() error {
 								Name:            fmt.Sprintf("csi-%splugin", r.driverType),
 								Image:           r.images["plugin"],
 								ImagePullPolicy: imagePullPolicy,
-								SecurityContext: securityContext,
+								SecurityContext: logRotateSecurityContext,
 								Args: utils.DeleteZeroValues(
 									[]string{
 										utils.TypeContainerArg(string(r.driverType)),
@@ -727,7 +738,7 @@ func (r *driverReconcile) reconcileControllerPluginDeployment() error {
 								Name:            "csi-addons",
 								Image:           r.images["addons"],
 								ImagePullPolicy: imagePullPolicy,
-								SecurityContext: securityContext,
+								SecurityContext: logRotateSecurityContext,
 								Args: utils.DeleteZeroValues(
 									append(
 										slices.Clone(leaderElectionSettingsArg),
@@ -832,8 +843,8 @@ func (r *driverReconcile) reconcileControllerPluginDeployment() error {
 								Image:           r.images["plugin"],
 								ImagePullPolicy: imagePullPolicy,
 								Resources:       resources,
-								Command:         []string{"/bin/bash", "-c"},
-								Args:            []string{`while true; do logrotate --verbose /logrotate-config/csi; sleep 15m; done`},
+								SecurityContext: logRotateSecurityContext,
+								Command:         []string{"/bin/bash", "-c", logRotateCmd},
 								VolumeMounts: []corev1.VolumeMount{
 									utils.LogsDirVolumeMount,
 									utils.LogRotateDirVolumeMount,
@@ -866,6 +877,7 @@ func (r *driverReconcile) reconcileControllerPluginDeployment() error {
 								utils.KmsConfigVolume(&r.driver.Spec.Encryption.ConfigMapRef))
 						}
 						if logRotationEnabled {
+							logHostPath := cmp.Or(logRotationSpec.LogHostPath, defaultLogHostPath)
 							volumes = append(
 								volumes,
 								utils.LogsDirVolume(logHostPath, deploy.Name),
@@ -913,7 +925,6 @@ func (r *driverReconcile) reconcileNodePluginDeamonSet() error {
 		topology := r.isRdbDriver() && pluginSpec.Topology != nil
 		domainLabels := cmp.Or(pluginSpec.Topology, &csiv1a1.TopologySpec{}).DomainLabels
 		logRotationSpec := cmp.Or(r.driver.Spec.Log, &csiv1a1.LogSpec{}).Rotation
-		logHostPath := cmp.Or(logRotationSpec.LogHostPath, defaultLogHostPath)
 		logRotationEnabled := logRotationSpec != nil
 
 		daemonSet.Spec = appsv1.DaemonSetSpec{
@@ -1162,8 +1173,13 @@ func (r *driverReconcile) reconcileNodePluginDeamonSet() error {
 								Image:           r.images["plugin"],
 								ImagePullPolicy: imagePullPolicy,
 								Resources:       resources,
-								Command:         []string{"/bin/bash", "-c"},
-								Args:            []string{`while true; do logrotate --verbose /logrotate-config/csi; sleep 15m; done`},
+								SecurityContext: &corev1.SecurityContext{
+									Privileged: ptr.To(true),
+									Capabilities: &corev1.Capabilities{
+										Drop: []corev1.Capability{"All"},
+									},
+								},
+								Command: []string{"/bin/bash", "-c", logRotateCmd},
 								VolumeMounts: []corev1.VolumeMount{
 									utils.LogsDirVolumeMount,
 									utils.LogRotateDirVolumeMount,
@@ -1212,6 +1228,7 @@ func (r *driverReconcile) reconcileNodePluginDeamonSet() error {
 							)
 						}
 						if logRotationEnabled {
+							logHostPath := cmp.Or(logRotationSpec.LogHostPath, defaultLogHostPath)
 							volumes = append(
 								volumes,
 								utils.LogsDirVolume(logHostPath, daemonSet.Name),
@@ -1460,6 +1477,9 @@ func mergeDriverSpecs(dest, src *csiv1a1.DriverSpec) {
 			}
 			if dest.Replicas == nil {
 				dest.Replicas = src.Replicas
+			}
+			if dest.Privileged == nil {
+				dest.Privileged = src.Privileged
 			}
 			if dest.Resources.Attacher == nil {
 				dest.Resources.Attacher = src.Resources.Attacher
--- a/internal/utils/csi.go
+++ b/internal/utils/csi.go
@@ -37,6 +37,9 @@ const (
 
 	CsiConfigMapConfigKey  = "config.json"
 	CsiConfigMapMappingKey = "cluster-mapping.json"
+
+	logsDirVolumeName      = "logs-dir"
+	logRotateDirVolumeName = "log-rotate-dir"
 )
 
 // Ceph CSI common volumes
@@ -182,6 +185,29 @@ func PodsMountDirVolume(kubeletDirPath string) corev1.Volume {
 		},
 	}
 }
+func LogsDirVolume(logHostPath, pluginName string) corev1.Volume {
+	return corev1.Volume{
+		Name: logsDirVolumeName,
+		VolumeSource: corev1.VolumeSource{
+			HostPath: &corev1.HostPathVolumeSource{
+				Path: fmt.Sprintf("%s/%s", logHostPath, pluginName),
+				Type: ptr.To(corev1.HostPathDirectoryOrCreate),
+			},
+		},
+	}
+}
+func LogRotateDirVolumeName(driverName string) corev1.Volume {
+	return corev1.Volume{
+		Name: logRotateDirVolumeName,
+		VolumeSource: corev1.VolumeSource{
+			ConfigMap: &corev1.ConfigMapVolumeSource{
+				LocalObjectReference: corev1.LocalObjectReference{
+					Name: LogRotateConfigMapName(driverName),
+				},
+			},
+		},
+	}
+}
 
 // Ceph CSI common volume Mounts
 var SocketDirVolumeMount = corev1.VolumeMount{
@@ -235,6 +261,14 @@ var EtcSelinuxVolumeMount = corev1.VolumeMount{
 	Name:      EtcSelinuxVolume.Name,
 	MountPath: "/etc/selinux",
 	ReadOnly:  true,
+}
+var LogsDirVolumeMount = corev1.VolumeMount{
+	Name:      logsDirVolumeName,
+	MountPath: "/csi-logs",
+}
+var LogRotateDirVolumeMount = corev1.VolumeMount{
+	Name:      logRotateDirVolumeName,
+	MountPath: "/logrotate-config",
 }
 
 func PodsMountDirVolumeMount(kubletDirPath string) corev1.VolumeMount {
@@ -335,9 +369,14 @@ var ImmediateTopologyContainerArg = "--immediate-topology=false"
 var RecoverVolumeExpansionFailureContainerArg = "--feature-gates=RecoverVolumeExpansionFailure=true"
 var EnableVolumeGroupSnapshotsContainerArg = "--enable-volume-group-snapshots=true"
 var ForceCephKernelClientContainerArg = "--forcecephkernelclient=true"
+var LogToStdErrContainerArg = "--logtostderr=false"
+var AlsoLogToStdErrContainerArg = "--alsologtostderr=true"
 
 func LogVerbosityContainerArg(level int) string {
 	return fmt.Sprintf("--v=%d", Clamp(level, 0, 5))
+}
+func LogFileContainerArg(containerName string) string {
+	return fmt.Sprintf("--log_file=csi-logs/%s.log", containerName)
 }
 func TypeContainerArg(t string) string {
 	switch t {
@@ -379,6 +418,9 @@ func KubeletRegistrationPathContainerArg(kubeletDirPath string, driverName strin
 }
 func StagingPathContainerArg(kubeletDirPath string) string {
 	return fmt.Sprintf("--stagingpath=%s/plugins/kubernetes.io/csi/", kubeletDirPath)
+}
+func LogRotateConfigMapName(driverName string) string {
+	return fmt.Sprintf("%s-logrotate-config", driverName)
 }
 func KernelMountOptionsContainerArg(options map[string]string) string {
 	return If(
--- a/config/crd/bases/csi.ceph.io_drivers.yaml
+++ b/config/crd/bases/csi.ceph.io_drivers.yaml
@@ -3542,11 +3542,14 @@ spec:
                     description: log rotation for csi pods
                     properties:
                       logHostPath:
-                        description: LogHostPath is the prefix directory path for
-                          the csi log files
+                        description: |-
+                          LogHostPath is the prefix directory path for the csi log files
+                          Default to /var/lib/cephcsi
                         type: string
                       maxFiles:
-                        description: MaxFiles is the number of logrtoate files
+                        description: |-
+                          MaxFiles is the number of logrtoate files
+                          Default to 7
                         type: integer
                       maxLogSize:
                         anyOf:
@@ -3565,6 +3568,9 @@ spec:
                         - monthly
                         type: string
                     type: object
+                    x-kubernetes-validations:
+                    - message: Either maxLogSize or periodicity must be set
+                      rule: (has(self.maxLogSize)) || (has(self.periodicity))
                   verbosity:
                     description: |-
                       Log verbosity level for driver pods,
--- a/config/crd/bases/csi.ceph.io_operatorconfigs.yaml
+++ b/config/crd/bases/csi.ceph.io_operatorconfigs.yaml
@@ -3581,11 +3581,14 @@ spec:
                         description: log rotation for csi pods
                         properties:
                           logHostPath:
-                            description: LogHostPath is the prefix directory path
-                              for the csi log files
+                            description: |-
+                              LogHostPath is the prefix directory path for the csi log files
+                              Default to /var/lib/cephcsi
                             type: string
                           maxFiles:
-                            description: MaxFiles is the number of logrtoate files
+                            description: |-
+                              MaxFiles is the number of logrtoate files
+                              Default to 7
                             type: integer
                           maxLogSize:
                             anyOf:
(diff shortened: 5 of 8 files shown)


parth-gr:
@nb-ohad @Madhu-1 ^^

internal/controller/driver_controller.go
```diff
@@ -477,6 +479,9 @@ func (r *driverReconcile) reconcileControllerPluginDeployment() error {
 										utils.CsiAddonsEndpointContainerArg,
 										"",
 									),
+									utils.If(r.logRotation(), utils.LogRotationSpecArg(r.driver.Spec.Log.Rotation.LogHostPath, deploy.Name, fmt.Sprintf("csi-%splugin", r.driverType)), ""),
```
nb-ohad, line 482: @parth-gr 
This will panic, unlike a built-in `if` all the arguments are first evaluated and only then passed into the `util. If` function (this is the normal behavior for any function call in most languages). This means that the expression `r.driver.Spec.Log.Rotation.LogHostPath` will try to be evaluated no matter what the result of `r.logRotation()` is. 
This in turn will result in a nil de-reference which will panic.

internal/controller/driver_controller.go
```diff
@@ -477,6 +479,9 @@ func (r *driverReconcile) reconcileControllerPluginDeployment() error {
...
+									utils.If(r.logRotation(), utils.LogToFile, ""),
+									utils.If(r.logRotation(), utils.LogToStderr, ""),
```
nb-ohad, line 484: These three arguments are repeated in many containers. Please calc them into a separate Array at the op of the function, where you can start with an empty slice then append if log rotation is enabled, then join/concat them into the existing argument list

internal/controller/driver_controller.go
```diff
@@ -499,6 +504,7 @@ func (r *driverReconcile) reconcileControllerPluginDeployment() error {
 										utils.LibModulesVolumeMount,
 										utils.KeysTmpDirVolumeMount,
 										utils.CsiConfigVolumeMount,
+										utils.If(r.logRotation(), utils.LogDirVolumeMount(r.driver.Spec.Log.Rotation.LogHostPath, deploy.Name), corev1.VolumeMount{}),
```
nb-ohad, line 507: A couple of issues here:
1. This also results in a panic, for the same reason specified in a previous comment
2. This will add an empty VolumeMount to the volume list if log rotation is not enabled which is undesirable behavior

For conditional volumes, please follow the example provided by the `KmsConfigVolumeMount` a couple of lines below.

internal/controller/driver_controller.go
```diff
@@ -461,6 +462,7 @@ func (r *driverReconcile) reconcileControllerPluginDeployment() error {
 								Name:            fmt.Sprintf("csi-%splugin", r.driverType),
 								Image:           r.images["plugin"],
 								ImagePullPolicy: imagePullPolicy,
+								SecurityContext: utils.If(r.logRotation(), utils.Privileged(), &corev1.SecurityContext{Privileged: ptr.To(false)}),
```
nb-ohad, line 465: You repeat the same `If` for SecurityContext on every/most container. 
In that case, please precalculate the value once above the deployment spec definition into a local variable and use that variable instead. You can take an example from the pre-calc for `imagePullPolicy`

internal/utils/csi.go
```diff
@@ -119,6 +121,27 @@ var EtcSelinuxVolume = corev1.Volume{
...
+
+func LogSidecarDirVolume(LogHostPath, CsiComponentName string) corev1.Volume {
+	return corev1.Volume{
+		Name: csiSidecarLogDirectoryVolumeName,
+		VolumeSource: corev1.VolumeSource{
+			EmptyDir: &corev1.EmptyDirVolumeSource{},
+		},
+	}
+}
+
```
nb-ohad, line 144: Please move these to after the definitions of the existing volume functions

internal/utils/csi.go
```diff
@@ -119,6 +121,27 @@ var EtcSelinuxVolume = corev1.Volume{
...
+		Name: csiLogDirectoryVolumeName,
+		VolumeSource: corev1.VolumeSource{
+			HostPath: &corev1.HostPathVolumeSource{
+				Path: fmt.Sprintf("%s/log/%s", LogHostPath, CsiComponentName),
+				Type: ptr.To(corev1.HostPathDirectoryOrCreate),
+			},
+		},
+	}
+}
+
```
nb-ohad, line 135:
```suggestion
```

internal/utils/csi.go
```diff
@@ -119,6 +121,27 @@ var EtcSelinuxVolume = corev1.Volume{
...
+func LogSidecarDirVolume(LogHostPath, CsiComponentName string) corev1.Volume {
```
nb-ohad, line 136: The arguments for this volume function are not used. 
Why is this a function and not a static var volume?

internal/utils/csi.go
```diff
@@ -28,11 +28,13 @@ const (
...
-	registrationVolumeName   = "registration-dir"
-	pluginDirVolumeName      = "plugin-dir"
-	podsMountDirVolumeName   = "pods-mount-dir"
-	pluginMountDirVolumeName = "plugin-mount-dir"
+	kmsConfigVolumeName              = "ceph-csi-kms-config"
+	registrationVolumeName           = "registration-dir"
+	pluginDirVolumeName              = "plugin-dir"
+	podsMountDirVolumeName           = "pods-mount-dir"
+	pluginMountDirVolumeName         = "plugin-mount-dir"
+	csiLogDirectoryVolumeName        = "csi-log"
```
nb-ohad, line 36: please make the var name match the exiting conventions:
So in this case, if the volume name is `csi-log` the const name should be `csiLogVolumeName`?

In addition, there is no need to use a `csi` prefix. Every volume in this file is a csi volume.
May I suggest `logsDirVolumeName = "logs-dir"`

internal/utils/csi.go
```diff
@@ -28,11 +28,13 @@ const (
...
+	csiSidecarLogDirectoryVolumeName = "csi-logs-logrotate"
```
nb-ohad, line 37: Similar to the above comment:
`logRotateDirVolumeName = "log-rotate-dir"`

internal/utils/csi.go
```diff
@@ -119,6 +121,27 @@ var EtcSelinuxVolume = corev1.Volume{
 	},
 }
 
+func LogDirVolume(LogHostPath, CsiComponentName string) corev1.Volume {
```
nb-ohad, line 124: Why do we need the csi component name as part of the volume name?
I don't see a need to mount a separate volume for each component.

You might want to use different folders inside the same mount, but not different volumes

internal/utils/csi.go
```diff
@@ -331,10 +366,15 @@ var TopologyContainerArg = "--feature-gates=Topology=true"
 var RecoverVolumeExpansionFailureContainerArg = "--feature-gates=RecoverVolumeExpansionFailure=true"
 var EnableVolumeGroupSnapshotsContainerArg = "--enable-volume-group-snapshots=true"
 var ForceCephKernelClientContainerArg = "--forcecephkernelclient=true"
+var LogToFile = "--logtostderr=false"
+var LogToStderr = "--alsologtostderr=true"
```
nb-ohad, line 370: Argument const names should match the argument string and should always end in `ContainerArg`. This is done for each of readability when looking at the definition of the container arguments. 

```
var LogToStdErrContainerArg = "--logtostderr=false"
var AlsoLogToStdErrContainerArg = "--alsologtostderr=true"
```

internal/utils/csi.go
```diff
@@ -331,10 +366,15 @@ var TopologyContainerArg = "--feature-gates=Topology=true"
...
 
 func LogLevelContainerArg(level int) string {
 	return fmt.Sprintf("--v=%d", Clamp(level, 0, 5))
 }
+func LogRotationSpecArg(LogHostPath, CsiComponentName, containerName string) string {
```
nb-ohad, line 375: Name should be `LogFileContainerArg` to match the arg name and the `ContainerArg` suffix

internal/utils/csi.go
```diff
@@ -390,3 +430,9 @@ func FuseMountOptionsContainerArg(options map[string]string) string {
 		"",
 	)
 }
+
+func Privileged() *corev1.SecurityContext {
+	return &corev1.SecurityContext{
+		Privileged: ptr.To(true),
+	}
+}
```
nb-ohad, line 438: Looking at the call sites, it does not make any sense to have a util that creates the privileged security context but not the unprivileged security context. Let's remove this func and apply the strategy I mentioned in a previous comment.

internal/controller/driver_controller.go
```diff
@@ -1259,3 +1299,55 @@ func mergeDriverSpecs(dest, src *csiv1a1.DriverSpec) {
 		dest.CephFsClientType = src.CephFsClientType
 	}
 }
+
+func (r *driverReconcile) logrRotateContainer(imagePullPolicy corev1.PullPolicy, csiComponentName string, resources *corev1.ResourceRequirements) corev1.Container {
```
nb-ohad, line 1303: Please have the container definition inline like any other container already defined.

internal/controller/driver_controller.go
```diff
@@ -1259,3 +1299,55 @@ func mergeDriverSpecs(dest, src *csiv1a1.DriverSpec) {
...
+LOG_ROTATE_CEPH_CSI_FILE="$CsiLogHostPath"/logrotate-config/"$CsiComponentName"/csi
+if [ "$LOG_MAX_SIZE" != "0" ]; then
+  sed --in-place "4i \ \ \ \ maxsize $LOG_MAX_SIZE" "$LOG_ROTATE_CEPH_CSI_FILE"
+fi
+
+while true; do
+  logrotate --verbose "$LOG_ROTATE_CEPH_CSI_FILE"
+  sleep 15m
+done
+`
```
nb-ohad, line 1353: Why do we have an inline bash script in our go code? 
If you need to have a log rotate sidecar build it as an image so it can be controlled by all of the image control mechanisms we already have

internal/controller/driver_controller.go
```diff
@@ -1053,6 +1083,10 @@ func (r *driverReconcile) isNfsDriver() bool {
 	return r.driverType == NfsDriverType
 }
 
+func (r *driverReconcile) logRotation() bool {
+	return r.driver.Spec.Log.Rotation != nil
+}
```
nb-ohad, line 1088: See the previous comment about pre-calculating the argument.
It should make this function redundant

internal/controller/driver_controller.go
```diff
@@ -967,6 +995,8 @@ func (r *driverReconcile) reconcileNodePluginDeamonSet() error {
 							utils.PluginMountDirVolume(pluginSpec.KubeletDirPath),
 							utils.PodsMountDirVolume(pluginSpec.KubeletDirPath),
 							utils.RegistrationDirVolume(pluginSpec.KubeletDirPath),
+							utils.If(r.logRotation(), utils.LogDirVolume(r.driver.Spec.Log.Rotation.LogHostPath, daemonSet.Name), corev1.Volume{}),
+							utils.If(r.logRotation(), utils.LogSidecarDirVolume(r.driver.Spec.Log.Rotation.LogHostPath, daemonSet.Name), corev1.Volume{}),
```
nb-ohad, line 999: You should not use `utils.If` and an empty volume definition in the volume list.
See a previous comment for the reasoning

internal/controller/driver_controller.go
```diff
@@ -946,6 +970,10 @@ func (r *driverReconcile) reconcileNodePluginDeamonSet() error {
 								})
 							}
 						}
+						// CSI Logrotate Container
+						if r.driver.Spec.Log.Rotation != nil {
+							containers = append(containers, r.logrRotateContainer(imagePullPolicy, daemonSet.Name, r.driver.Spec.NodePlugin.Resources.LogRotator))
+						}
```
nb-ohad, line 976: Please inline the log rotate container definition

internal/controller/driver_controller.go
```diff
@@ -907,6 +930,7 @@ func (r *driverReconcile) reconcileNodePluginDeamonSet() error {
 								},
 								VolumeMounts: []corev1.VolumeMount{
 									utils.PluginDirVolumeMount,
+									utils.If(r.logRotation(), utils.LogDirVolumeMount(r.driver.Spec.Log.Rotation.LogHostPath, daemonSet.Name), corev1.VolumeMount{}),
```
nb-ohad, line 933: You should not use `utils.If` and an empty volume mount definition in the volume list.
See a previous comment for the reasoning

internal/controller/driver_controller.go
```diff
@@ -828,6 +847,7 @@ func (r *driverReconcile) reconcileNodePluginDeamonSet() error {
 										utils.PluginDirVolumeMount,
 										utils.PluginMountDirVolumeMount(kubeletDirPath),
 										utils.PodsMountDirVolumeMount(kubeletDirPath),
+										utils.If(r.logRotation(), utils.LogDirVolumeMount(r.driver.Spec.Log.Rotation.LogHostPath, daemonSet.Name), corev1.VolumeMount{}),
```
nb-ohad, line 850: You should not use `utils.If` and an empty volume mount definition in the volume list.
See a previous comment for the reasoning

nb-ohad (changes requested):

internal/utils/csi.go
parth-gr: I see the other volume functions are just below it

internal/controller/driver_controller.go
parth-gr: I dont see any other common container function

internal/controller/driver_controller.go
parth-gr: this isn't more readable, what is your suggestion can you elaborate

internal/controller/driver_controller.go
BlaineEXE: Rook has had very good success using inline bash scripts to help tie functionality together. 

For one, it allows using any desired image, which helps with avoiding unnecessary pod restarts. For example, using the ceph-csi-operator container here means that the deployment/daemonset needs to be restarted with every new operator version. In contrast, if the ceph-csi image is used here, the restart is only needed when that image is already changing.

For a second, having dedicated images for stuff like this ends up being cumbersome and limits developer flexibility. When simple scripts are used, code maintenance goes a lot more quickly than in the past where Rook was using golang daemons for everything. It also means that any potential script bugs can be fixed in the operator and don't need to be fixed on a per-CSI-release basis.

If you are proposing to add this functionality to the ceph-csi image natively, we can probably work together to help implement that, but I still think the inline script provides good flexibility and transparency, and I think that is a consideration worth making.

But if you are proposing to create an entirely new image, I would push back to say that I don't think Parth should be the one managing that complexity.

BlaineEXE: @nb-ohad this has been a topic of discussion in the Rook huddle for a couple days. I want to make sure my prior comment didn't get lost in the amount of other comments here. What are your and @Madhu-1 's thoughts on this?

internal/controller/driver_controller.go
```diff
@@ -436,6 +437,30 @@ func (r *driverReconcile) reconcileControllerPluginDeployment() error {
...
+				utils.LogFileContainerArg(fmt.Sprintf("csi-%splugin", r.driverType)),
+			}
+			logRotationCsiAddonsSidecarContainerArgs = []string{
+				utils.LogToStdErrContainerArg,
+				utils.AlsoLogToStdErrContainerArg,
+				utils.LogFileContainerArg("csi-addons"),
+			}
+			securityContext = &corev1.SecurityContext{
+				Privileged: ptr.To(true),
+			}
```
Madhu-1, line 461: Possible to move this to helper to avoid duplication?

internal/utils/csi.go
```diff
@@ -178,6 +186,17 @@ func PodsMountDirVolume(kubeletDirPath string) corev1.Volume {
 		},
 	}
 }
+func LogsDirVolume(LogHostPath, pluginName string) corev1.Volume {
```
Madhu-1, line 189: `LogHostPath` to `logHostPath`

Madhu-1 (changes requested):

internal/controller/driver_controller.go
```diff
@@ -1259,3 +1381,34 @@ func mergeDriverSpecs(dest, src *csiv1a1.DriverSpec) {
 		dest.CephFsClientType = src.CephFsClientType
 	}
 }
+
+var cronLogRotate = `
+echo "Starting the csi-logrotate-sidecar"
+PERIODICITY=%s
```
parth-gr, line 1387: @nb-ohad can you please summarise the security gaps here, for better clarity and future understanding,

What I see here,
1) we take inputs to some variables (anything can be taken as input moreover we can validate)
2) We create a file with those variables
3) We run `logrotate --verbose log-file`, now in 3rd step if there is an irrelevant input to the file, the logrotate command will just fail as it has its own validation.

Now if it's a container image doing these tasks, I see that the underlying logic would be hidden under the image, and the rest will be the same.

internal/controller/driver_controller.go
parth-gr: Done

parth-gr:
Testing:
```
[rider@localhost rbac]$ oc exec -it test.rbd.csi.ceph.com-nodeplugin-tg25d -n ceph-csi-operator-system  -c log-collector sh
kubectl exec [POD] [COMMAND] is DEPRECATED and will be removed in a future version. Use kubectl exec [POD] -- [COMMAND] instead.
sh-4.4# 
sh-4.4# ls
bin  csi-logs  dev  etc  home  lib  lib64  logrotate-config  lost+found  media	mnt  opt  proc	root  run  sbin  srv  sys  tmp	usr  var
sh-4.4# cd csi-logs/
sh-4.4# ls 
csi-rbdplugin.log
sh-4.4# 
sh-4.4# cd ..
sh-4.4# cd logrotate-config/
sh-4.4# ls
csi
sh-4.4# cat csi 

/csi-logs/*.log {
	daily
	missingok
	rotate 5
	compress
	copytruncate
	notifempty
    maxsize 500M
}sh-4.4# cd ..
sh-4.4# cd csi-logs/
sh-4.4# ls
csi-rbdplugin.log
sh-4.4# truncate -s 600M csi-rbdplugin.log 
sh-4.4# ls -lh
total 4.0K
-rw-r--r-- 1 root root 600M Aug  7 18:47 csi-rbdplugin.log
sh-4.4# 
sh-4.4# 
sh-4.4# cd ..
sh-4.4# cd csi-logs/
sh-4.4# logrotate --verbose csi               
error: cannot stat csi: No such file or directory
Reading state from file: /var/lib/logrotate/logrotate.status
Allocating hash table for state file, size 64 entries

Handling 0 logs
sh-4.4# 
sh-4.4# 
sh-4.4# ls
csi-rbdplugin.log
sh-4.4# cd ..
sh-4.4# cd logrotate-config/
sh-4.4# logrotate --verbose csi 
reading config file csi
Reading state from file: /var/lib/logrotate/logrotate.status
Allocating hash table for state file, size 64 entries

Handling 1 logs

rotating pattern: /csi-logs/*.log  after 1 days (5 rotations)
empty log files are not rotated, log files >= 524288000 are rotated earlier, old logs are removed
considering log /csi-logs/csi-rbdplugin.log
Creating new state
  Now: 2024-08-07 18:48
  Last rotated at 2024-08-07 18:00
  log needs rotating
rotating log /csi-logs/csi-rbdplugin.log, log->rotateCount is 5
dateext suffix '-20240807'
glob pattern '-[0-9][0-9][0-9][0-9][0-9][0-9][0-9][0-9]'
renaming /csi-logs/csi-rbdplugin.log.5.gz to /csi-logs/csi-rbdplugin.log.6.gz (rotatecount 5, logstart 1, i 5), 
old log /csi-logs/csi-rbdplugin.log.5.gz does not exist
renaming /csi-logs/csi-rbdplugin.log.4.gz to /csi-logs/csi-rbdplugin.log.5.gz (rotatecount 5, logstart 1, i 4), 
old log /csi-logs/csi-rbdplugin.log.4.gz does not exist
renaming /csi-logs/csi-rbdplugin.log.3.gz to /csi-logs/csi-rbdplugin.log.4.gz (rotatecount 5, logstart 1, i 3), 
old log /csi-logs/csi-rbdplugin.log.3.gz does not exist
renaming /csi-logs/csi-rbdplugin.log.2.gz to /csi-logs/csi-rbdplugin.log.3.gz (rotatecount 5, logstart 1, i 2), 
old log /csi-logs/csi-rbdplugin.log.2.gz does not exist
renaming /csi-logs/csi-rbdplugin.log.1.gz to /csi-logs/csi-rbdplugin.log.2.gz (rotatecount 5, logstart 1, i 1), 
old log /csi-logs/csi-rbdplugin.log.1.gz does not exist
renaming /csi-logs/csi-rbdplugin.log.0.gz to /csi-logs/csi-rbdplugin.log.1.gz (rotatecount 5, logstart 1, i 0), 
old log /csi-logs/csi-rbdplugin.log.0.gz does not exist
log /csi-logs/csi-rbdplugin.log.6.gz doesn't exist -- won't try to dispose of it
copying /csi-logs/csi-rbdplugin.log to /csi-logs/csi-rbdplugin.log.1
truncating /csi-logs/csi-rbdplugin.log
compressing log with: /bin/gzip


sh-4.4# 
sh-4.4# 
sh-4.4# cd ..
sh-4.4# cd csi-logs/
sh-4.4# ls
csi-rbdplugin.log  csi-rbdplugin.log.1.gz

```

internal/controller/driver_controller.go
nb-ohad: @BlaineEXE Regardless of the reasons, which we had several discussions on offline and which are mainly security concerns, we decided to shift to having the log-rotate config map be created by the operator and mounted into the pod using a volume. That by itself, make any kind of injected scripting unnecessary for this specific use case

internal/controller/driver_controller.go
```diff
@@ -462,6 +515,23 @@ func (r *driverReconcile) reconcileControllerPluginDeployment() error {
 			utils.LeaderElectionRetryPeriodContainerArg(leaderElectionSpec.RetryPeriod),
 		}
 
+		logRotator := &csiv1a1.LogRotationSpec{}
+		if r.driver.Spec.Log != nil {
+			logRotator = r.driver.Spec.Log.Rotation
+		}
```
nb-ohad, line 521:
```suggestion
		logRotationSpec := cmp.Or(r.driver.Spec.Log, csiv1a1.LogSpec{}).Rotation
```

internal/controller/driver_controller.go
```diff
@@ -462,6 +515,23 @@ func (r *driverReconcile) reconcileControllerPluginDeployment() error {
...
+		var logHostPath string
+		securityContext := &corev1.SecurityContext{Privileged: ptr.To(false)}
+		// choosing a different approach as need to check nil values
+		if logRotator != nil {
+			logRotationContainerArgs, logRotationCsiAddonsSidecarContainerArgs = r.setLogrotateDefaultConfig()
+			securityContext = &corev1.SecurityContext{
+				Privileged: ptr.To(true),
+			}
+			logHostPath = cmp.Or(logRotator.LogHostPath, defaultLogHostPath)
+		}
```
nb-ohad, line 533: Please inline these as we do with any other container arguments

internal/controller/driver_controller.go
```diff
@@ -487,8 +557,10 @@ func (r *driverReconcile) reconcileControllerPluginDeployment() error {
 								Name:            fmt.Sprintf("csi-%splugin", r.driverType),
 								Image:           r.images["plugin"],
 								ImagePullPolicy: imagePullPolicy,
+								SecurityContext: securityContext,
 								Args: utils.DeleteZeroValues(
-									[]string{
+									append(
+										slices.Clone(logRotationContainerArgs),
```
nb-ohad, line 563: This is being used only once, there is no need to declare it as a slice above and append it here. Just have the specific settings injected here instead

internal/controller/driver_controller.go
```diff
@@ -528,6 +599,9 @@ func (r *driverReconcile) reconcileControllerPluginDeployment() error {
 										utils.KeysTmpDirVolumeMount,
 										utils.CsiConfigVolumeMount,
 									)
+									if logRotator != nil {
+										mounts = append(mounts, utils.LogsDirVolumeMount)
+									}
```
nb-ohad, line 604: Can you please move this to be the last volume added?

internal/controller/driver_controller.go
```diff
@@ -654,18 +728,20 @@ func (r *driverReconcile) reconcileControllerPluginDeployment() error {
...
 								Image:           r.images["addons"],
 								ImagePullPolicy: imagePullPolicy,
-								Args: utils.DeleteZeroValues(
-									append(
-										slices.Clone(leaderElectionSettingsArg),
+								SecurityContext: securityContext,
+								Args: utils.DeleteZeroValues(slices.Concat(
+									leaderElectionSettingsArg,
+									logRotationCsiAddonsSidecarContainerArgs,
+									[]string{
```
nb-ohad, line 735: This large change is unnecessary.
These log rotation container args are only used once, there is no need to declare it as a slice above and append it here. Just have the specific settings injected here instead

internal/controller/driver_controller.go
```diff
@@ -741,7 +823,26 @@ func (r *driverReconcile) reconcileControllerPluginDeployment() error {
 								),
 							})
 						}
-
+						// CSI Logrotate Container
+						if logRotator != nil {
+							resources := ptr.Deref(pluginSpec.Resources.LogRotator, corev1.ResourceRequirements{})
+							containers = append(containers, corev1.Container{
+								Name:            "log-collector",
```
nb-ohad, line 830: Why collector and not rotator? I don't see an need to change established terminology here
```suggestion
								Name:            "log-rotator",
```

internal/controller/driver_controller.go
```diff
@@ -741,7 +823,26 @@ func (r *driverReconcile) reconcileControllerPluginDeployment() error {
...
+								Image:           r.images["plugin"],
+								ImagePullPolicy: imagePullPolicy,
+								Resources:       resources,
+								Command: []string{
+									"sh", "-c", `while true; do 
+								    	sleep 15m
+										logrotate --verbose /logrotate-config/csi
+									done`,
```
nb-ohad, line 838: Please explain why we need this elaborate command and not just the `logrotate` line.

internal/controller/driver_controller.go
```diff
@@ -762,6 +863,9 @@ func (r *driverReconcile) reconcileControllerPluginDeployment() error {
 							utils.OidcTokenVolume,
 							utils.CsiConfigVolume,
 						)
+						if logRotator != nil {
+							volumes = append(volumes, utils.LogsDirVolume(logHostPath, deploy.Name), utils.LogRotateDirVolumeName(r.driver.Name))
+						}
```
nb-ohad, line 868: Can you please move this to be the last volume added?

internal/controller/driver_controller.go
```diff
@@ -807,6 +911,18 @@ func (r *driverReconcile) reconcileNodePluginDeamonSet() error {
...
+			logRotator = r.driver.Spec.Log.Rotation
+		}
+		var logRotationContainerArgs []string
+		var logRotationCsiAddonsSidecarContainerArgs []string
+		var logHostPath string
+		// choosing a different approach as need to check nil values
+		if logRotator != nil {
+			logRotationContainerArgs, logRotationCsiAddonsSidecarContainerArgs = r.setLogrotateDefaultConfig()
+			logHostPath = cmp.Or(logRotator.LogHostPath, defaultLogHostPath)
+		}
```
nb-ohad, line 925: Please see similar previous comments on a similar code block above

internal/controller/driver_controller.go
```diff
@@ -850,7 +966,8 @@ func (r *driverReconcile) reconcileNodePluginDeamonSet() error {
 									AllowPrivilegeEscalation: ptr.To(true),
 								},
 								Args: utils.DeleteZeroValues(
-									[]string{
+									append(
+										slices.Clone(logRotationContainerArgs),
```
nb-ohad, line 970: Please see similar previous comments on a similar code block above

internal/controller/driver_controller.go
```diff
@@ -963,17 +1082,18 @@ func (r *driverReconcile) reconcileNodePluginDeamonSet() error {
 									},
 								},
 								Args: utils.DeleteZeroValues(
-									[]string{
+									append(
+										slices.Clone(logRotationCsiAddonsSidecarContainerArgs),
```
nb-ohad, line 1086: Please see similar previous comments on a similar code block above

internal/controller/driver_controller.go
```diff
@@ -1026,6 +1152,26 @@ func (r *driverReconcile) reconcileNodePluginDeamonSet() error {
...
+								Name:            "log-collector",
+								Image:           r.images["plugin"],
+								ImagePullPolicy: imagePullPolicy,
+								Resources:       resources,
+								Command: []string{
+									"sh", "-c", `while true; do 
+								    	sleep 15m
+										logrotate --verbose /logrotate-config/csi
+									done`,
+								},
```
nb-ohad, line 1168: Please see similar previous comments on a similar code block above

internal/controller/driver_controller.go
```diff
@@ -1049,6 +1195,9 @@ func (r *driverReconcile) reconcileNodePluginDeamonSet() error {
 							utils.PodsMountDirVolume(kubeletDirPath),
 							utils.RegistrationDirVolume(kubeletDirPath),
 						)
+						if logRotator != nil {
+							volumes = append(volumes, utils.LogsDirVolume(logHostPath, daemonSet.Name), utils.LogRotateDirVolumeName(r.driver.Name))
+						}
```
nb-ohad, line 1200: Can you please move this to be the last volume added?
And make it a multi-line statement (the line is very long)

internal/controller/driver_controller.go
```diff
@@ -1182,6 +1331,20 @@ func logCreateOrUpdateResult(
...
+		utils.AlsoLogToStdErrContainerArg,
+		utils.LogFileContainerArg(fmt.Sprintf("csi-%splugin", r.driverType)),
+	}
+	logRotationCsiAddonsSidecarContainerArgs := []string{
+		utils.LogToStdErrContainerArg,
+		utils.AlsoLogToStdErrContainerArg,
+		utils.LogFileContainerArg("csi-addons"),
+	}
+	return logRotationContainerArgs, logRotationCsiAddonsSidecarContainerArgs
+}
```
nb-ohad, line 1346: There is no need for default configuration.

nb-ohad (changes requested):

api/v1alpha1/driver_types.go
```diff
@@ -35,6 +35,7 @@ const (
 
 type LogRotationSpec struct {
 	// MaxFiles is the number of logrtoate files
+	// Default to 7
```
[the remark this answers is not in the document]
parth-gr: As discussed this can be set, and we will check for maxsize and periodicity

internal/controller/driver_controller.go
```diff
@@ -230,6 +230,7 @@ func (r *driverReconcile) reconcile() error {
 	// the desired state defined on the driver object
 	errChan := utils.RunConcurrently(
 		r.reconcileCsiConfigMap,
+		r.reconcileLogrotateConfigMap,
```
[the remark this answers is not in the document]
parth-gr: Actually logrotate is actually a single word, but changed it for now

internal/controller/driver_controller.go
```diff
@@ -327,6 +328,58 @@ func (r *driverReconcile) LoadAndValidateDesiredState() error {
 	return nil
 }
 
+func (r *driverReconcile) reconcileLogrotateConfigMap() error {
+	logRotator := &csiv1a1.LogRotationSpec{}
+	if r.driver.Spec.Log != nil {
+		logRotator = r.driver.Spec.Log.Rotation
+	}
+	if logRotator == nil {
+		return nil
```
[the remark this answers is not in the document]
parth-gr: Was this also needed in `reconcileCsiConfigMap`??

internal/controller/driver_controller.go
parth-gr: We need to run the container every 15 mins
And so we need a while loop running `logrotate command`

As it is container logroate cron job wont work so we need to explicitly run it

internal/utils/csi.go
```diff
@@ -181,6 +184,29 @@ func PodsMountDirVolume(kubeletDirPath string) corev1.Volume {
 		},
 	}
 }
+func LogsDirVolume(logHostPath, pluginName string) corev1.Volume {
```
nb-ohad, line 187: Our terminology does not use `pluginName`but `driverName`. Please align

nb-ohad (changes requested):

internal/controller/driver_controller.go
nb-ohad: First, there is a way to run cronjob in a container: https://github.com/Ekito/docker-cron
Second, even if you want to run manually, use `watch -n<time> <command>` instead of writing a bash loop

internal/controller/driver_controller.go
```diff
@@ -327,6 +328,66 @@ func (r *driverReconcile) LoadAndValidateDesiredState() error {
...
+				"\tcopytruncate",
+				"\tnotifempty",
+			}
+			if logRotationSpec.MaxFiles > 0 {
+				maxFiles := fmt.Sprintf("rotate %d", logRotationSpec.MaxFiles)
+				cronLogRotateSettings = append(cronLogRotateSettings, "\t"+maxFiles)
+			} else {
+				maxFiles := fmt.Sprintf("rotate %d", defaultLogRotateMaxFiles)
+				cronLogRotateSettings = append(cronLogRotateSettings, "\t"+maxFiles)
+			}
```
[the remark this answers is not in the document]
parth-gr: make sense

internal/controller/driver_controller.go
```diff
@@ -1182,6 +1339,20 @@ func logCreateOrUpdateResult(
 	}
 }
 
+func (r *driverReconcile) setLogrotateDefaultConfig() ([]string, []string) {
+	logRotationContainerArgs := []string{
+		utils.LogToStdErrContainerArg,
+		utils.AlsoLogToStdErrContainerArg,
+		utils.LogFileContainerArg(fmt.Sprintf("csi-%splugin", r.driverType)),
```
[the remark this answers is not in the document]
parth-gr: sure

internal/utils/csi.go
parth-gr: We are passing `utils.LogsDirVolume(logHostPath, daemonSet.Name)` not sure whats need to get change

api/v1alpha1/driver_types.go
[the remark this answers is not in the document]
nb-ohad: Ok, agreed

internal/controller/driver_controller.go
```diff
@@ -327,6 +328,61 @@ func (r *driverReconcile) LoadAndValidateDesiredState() error {
...
+
+	logRotationSpec := cmp.Or(r.driver.Spec.Log, &csiv1a1.LogSpec{}).Rotation
+	if logRotationSpec != nil {
+		opResult, err := ctrlutil.CreateOrUpdate(r.ctx, r.Client, logRotateConfigmap, func() error {
+			if _, err := utils.ToggleOwnerReference(true, logRotateConfigmap, &r.driver, r.Scheme); err != nil {
+				log.Error(err, "Failed adding an owner reference on the LogRotate config map")
+				return err
+			}
+			if logRotationSpec.Periodicity != "" && logRotationSpec.MaxLogSize.String() == "0" {
+				return fmt.Errorf("maxLogSize must be set when periodicity is set")
```
nb-ohad, line 347: Let's stick with simple well-defined error types. You can add a `log.Error` to capture everything else
```suggestion
				err := fmt.Error("Invalid Log.Rotation spec")
				log.Error(err, "Either \"maxLogSize\" or \"periodicity" fields must be set")
				return err 
```

internal/controller/driver_controller.go
```diff
@@ -327,6 +328,61 @@ func (r *driverReconcile) LoadAndValidateDesiredState() error {
...
+			cronLogRotateSettings := []string{
+				"\tmisslingok",
+				"\tcompress",
+				"\tcopytruncate",
+				"\tnotifempty",
+				fmt.Sprintf("\trotate %d", maxFiles),
+			}
+			if logRotationSpec.Periodicity != "" {
+				periodicity := string(logRotationSpec.Periodicity)
+				cronLogRotateSettings = append(cronLogRotateSettings, "\t"+periodicity)
```
nb-ohad, line 359: To align with the rest of the code, let's have the "\t" be precatenated as part of the value construction and not part of the append operation. 
```suggestion
				periodicity := "\t"+string(logRotationSpec.Periodicity)
				cronLogRotateSettings = append(cronLogRotateSettings, periodicity)
```

internal/controller/driver_controller.go
```diff
@@ -454,6 +510,13 @@ func (r *driverReconcile) reconcileControllerPluginDeployment() error {
 		logVerbosity := ptr.Deref(r.driver.Spec.Log, csiv1a1.LogSpec{}).Verbosity
 		forceKernelClient := r.isCephFsDriver() && r.driver.Spec.CephFsClientType == csiv1a1.KernelCephFsClient
 		snPolicy := cmp.Or(r.driver.Spec.SnapshotPolicy, csiv1a1.VolumeSnapshotSnapshotPolicy)
+		logRotationSpec := cmp.Or(r.driver.Spec.Log, &csiv1a1.LogSpec{}).Rotation
+		logHostPath := cmp.Or(logRotationSpec.LogHostPath, defaultLogHostPath)
+		securityContext := utils.If(
+			logRotationSpec != nil,
+			&corev1.SecurityContext{Privileged: ptr.To(true)},
+			&corev1.SecurityContext{Privileged: ptr.To(false)},
+		)
```
nb-ohad, line 519:
```suggestion
		securityContext := &corev1.SecurityContext{Privileged: ptr.To(logRotationSpec != nil)}
```

internal/controller/driver_controller.go
```diff
@@ -654,18 +723,20 @@ func (r *driverReconcile) reconcileControllerPluginDeployment() error {
...
 										utils.CsiAddonsNodeIdContainerArg,
 										utils.PodContainerArg,
 										utils.PodUidContainerArg,
 										utils.CsiAddonsAddressContainerArg,
 										utils.ControllerPortContainerArg,
 										utils.NamespaceContainerArg,
-									),
-								),
+									},
+								)),
```
nb-ohad: Please revert an inline the args as described in the last comment

internal/controller/driver_controller.go
```diff
@@ -850,7 +955,10 @@ func (r *driverReconcile) reconcileNodePluginDeamonSet() error {
 									AllowPrivilegeEscalation: ptr.To(true),
 								},
 								Args: utils.DeleteZeroValues(
-									[]string{
+									append(
+										slices.Clone(
+											utils.If(logRotationSpec != nil, utils.LogRotateContainerArgs(r.driver.Name, string(r.driverType)), []string{}),
+										),
```
nb-ohad: Same as the last comment

internal/controller/driver_controller.go
```diff
@@ -963,17 +1073,20 @@ func (r *driverReconcile) reconcileNodePluginDeamonSet() error {
 									},
 								},
 								Args: utils.DeleteZeroValues(
-									[]string{
+									append(
+										slices.Clone(
+											utils.If(logRotationSpec != nil, utils.LogRotationCsiAddonsSidecarContainerArgs(), []string{}),
+										),
 										utils.CsiAddonsNodeIdContainerArg,
+										utils.NodeIdContainerArg,
```
nb-ohad: Same as the last comment

nb-ohad (changes requested):

internal/utils/csi.go
nb-ohad: You should not use daemonSet.Name as the daemon set is not part of the desired state and there is no guarantee on its value. Please build the name from the driver's name.

This also means that `pluginName` is not the correct terminology

internal/controller/driver_controller.go
parth-gr: Much better way

internal/controller/driver_controller.go
```diff
@@ -504,6 +568,9 @@ func (r *driverReconcile) reconcileControllerPluginDeployment() error {
 											utils.CsiAddonsEndpointContainerArg,
 											"",
 										),
+										utils.If(logRotationSpec != nil, utils.LogToStdErrContainerArg, ""),
+										utils.If(logRotationSpec != nil, utils.AlsoLogToStdErrContainerArg, ""),
+										utils.If(logRotationSpec != nil, utils.LogFileContainerArg(fmt.Sprintf("%s-csi-%splugin", r.driver.Name, r.driverType)), ""),
```
nb-ohad, line 573: There is no point in having `csi` in the name, `csi` is already part of the driver name for every and all drivers
```suggestion
										utils.If(logRotationSpec != nil, utils.LogFileContainerArg(fmt.Sprintf("%s-csi-%splugin", r.driver.Name, r.driverType)), ""),
```

nb-ohad, line 573: You need to pass the container name as well

nb-ohad (changes requested):

internal/utils/csi.go
parth-gr: It is according to our logic 
```
Volume = <logHostPath>/<pluginName>
Volume Mount = /csi-logs 
Container Arg = /csi-logs/<container-name> 
```

I dont want it to break
Plus `daemonSet.Name` or `deploy.Name` is required, as to make different folders need to get created for daemonsets and deplouments pods

parth-gr: @nb-ohad

internal/controller/driver_controller.go
parth-gr: okay, even I think if we are making changes.
Users upgrading from rook logrotate to csi logrotate will see a change in names, should keep what we have finalised as per rook

parth-gr: it is actually a `drivername-containername` `%s-(csi-%splugin)`

internal/utils/csi.go
parth-gr: or I can do `Volume = <logHostPath>/<driverNmae>/<pluginName>`

internal/controller/driver_controller.go
parth-gr: Renamed

internal/controller/driver_controller.go
```diff
@@ -454,6 +512,10 @@ func (r *driverReconcile) reconcileControllerPluginDeployment() error {
 		logVerbosity := ptr.Deref(r.driver.Spec.Log, csiv1a1.LogSpec{}).Verbosity
 		forceKernelClient := r.isCephFsDriver() && r.driver.Spec.CephFsClientType == csiv1a1.KernelCephFsClient
 		snPolicy := cmp.Or(r.driver.Spec.SnapshotPolicy, csiv1a1.VolumeSnapshotSnapshotPolicy)
+		logRotationSpec := cmp.Or(r.driver.Spec.Log, &csiv1a1.LogSpec{}).Rotation
+		logHostPath := cmp.Or(logRotationSpec.LogHostPath, defaultLogHostPath)
+		logRotationEnabled := logRotationSpec != nil
+		securityContext := &corev1.SecurityContext{Privileged: ptr.To(logRotationEnabled)}
```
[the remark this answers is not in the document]
parth-gr: Got it

nb-ohad (approved):

nb-ohad:
@Madhu-1 This PR is ready to be merged, but merging is blocked on a change request that you have asked for that is already resolved. Can you please add your approved review so we can merge this?

internal/controller/driver_controller.go
```diff
@@ -327,6 +328,63 @@ func (r *driverReconcile) LoadAndValidateDesiredState() error {
...
+	log.Info("Reconciling logRotate configmap")
+
+	logRotationSpec := cmp.Or(r.driver.Spec.Log, &csiv1a1.LogSpec{}).Rotation
+	if logRotationSpec != nil {
+		opResult, err := ctrlutil.CreateOrUpdate(r.ctx, r.Client, logRotateConfigmap, func() error {
+			if _, err := utils.ToggleOwnerReference(true, logRotateConfigmap, &r.driver, r.Scheme); err != nil {
+				log.Error(err, "Failed adding an owner reference on the LogRotate config map")
+				return err
+			}
+			if logRotationSpec.Periodicity != "" && logRotationSpec.MaxLogSize.String() == "0" {
```
Madhu-1, line 346:
```suggestion
if logRotationSpec.Periodicity != "" && logRotationSpec.MaxLogSize.IsZero() {
```

use helper function IsZero to check if it's zero.

internal/controller/driver_controller.go
```diff
@@ -1067,6 +1209,13 @@ func (r *driverReconcile) reconcileNodePluginDeamonSet() error {
 								utils.OidcTokenVolume,
 							)
 						}
+						if logRotationEnabled {
+							volumes = append(
+								volumes,
+								utils.LogsDirVolume(logHostPath, daemonSet.Name),
```
Madhu-1, line 1215: what will happen for the case where we run two drivers in different namespaces that can get the same daemonset Name?

internal/controller/driver_controller.go
```diff
@@ -504,6 +567,13 @@ func (r *driverReconcile) reconcileControllerPluginDeployment() error {
 											utils.CsiAddonsEndpointContainerArg,
 											"",
 										),
+										utils.If(logRotationEnabled, utils.LogToStdErrContainerArg, ""),
+										utils.If(logRotationEnabled, utils.AlsoLogToStdErrContainerArg, ""),
+										utils.If(
+											logRotationEnabled,
+											utils.LogFileContainerArg(fmt.Sprintf("csi-%splugin", r.driverType)),
```
Madhu-1, line 574: what will happen if the controller-plugin and the node-plugin run on the same node? are they going to log into different files? i don't think so but need confirmation on this one.

Madhu-1 (changes requested):
@parth-gr can you please provide the test results

internal/controller/driver_controller.go
```diff
@@ -327,6 +328,58 @@ func (r *driverReconcile) LoadAndValidateDesiredState() error {
 	return nil
 }
 
+func (r *driverReconcile) reconcileLogrotateConfigMap() error {
+	logRotator := &csiv1a1.LogRotationSpec{}
+	if r.driver.Spec.Log != nil {
+		logRotator = r.driver.Spec.Log.Rotation
+	}
```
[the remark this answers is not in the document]
Madhu-1: This was old review, Marking it as resolved

internal/controller/driver_controller.go
parth-gr: daemonset and deployment Name is genrated using a generate function

`return fmt.Sprintf("%s-%s", r.driver.Name, suffix)`

The generate function appends the `drivername` as a prefix so no two `daemonset Name` can ever b same if its a differnt driver

internal/controller/driver_controller.go
parth-gr: FIle name would be same, but underlyning directory would be different ass per the volume,
`utils.LogsDirVolume(logHostPath, daemonSet.Name)`
`utils.LogsDirVolume(logHostPath, deploy.Name),`

internal/controller/driver_controller.go
```diff
@@ -487,6 +549,7 @@ func (r *driverReconcile) reconcileControllerPluginDeployment() error {
 								Name:            fmt.Sprintf("csi-%splugin", r.driverType),
 								Image:           r.images["plugin"],
 								ImagePullPolicy: imagePullPolicy,
+								SecurityContext: securityContext,
```
Madhu-1, line 552: we also need to drop all the capability as well

```
Capabilities: &corev1.Capabilities{
	Drop: []corev1.Capability{"All"},
	},
```

parth-gr: @Madhu-1 with that I remember we have a separate setting for running `provisioner` in privileged,
With the current change I am just looking to if the `logrotate is enabled`

So better would I should check if the logrotate is enabled as well as the spec setting is set we should enable securityContext right?

Madhu-1: if we are setting the securityContext we need to do the above as well to keep the minimal one. by default it wont drop all the capabilities.

parth-gr: Done

parth-gr:
Testing results:
```
[rider@localhost dist]$ kubectl get pod -nceph-csi-operator-system  -o wide
NAME                                                    READY   STATUS    RESTARTS   AGE     IP               NODE           NOMINATED NODE   READINESS GATES
ceph-csi-operator-controller-manager-5b85f8c9b8-jbcns   2/2     Running   0          28m     10.244.1.2       minikube-m02   <none>           <none>
test.rbd.csi.ceph.com-ctrlplugin-54b744ff6c-chht6       6/6     Running   0          27m     10.244.2.2       minikube-m03   <none>           <none>
test.rbd.csi.ceph.com-nodeplugin-dn4ww                  3/3     Running   0          27m     192.168.39.45    minikube-m03   <none>           <none>
test.rbd.csi.ceph.com-nodeplugin-s62ts                  3/3     Running   0          3m40s   192.168.39.232   minikube       <none>           <none>
test.rbd.csi.ceph.com-nodeplugin-vj2f2                  3/3     Running   0          27m     192.168.39.23    minikube-m02   <none>           <none>
[rider@localhost dist]$ 
[rider@localhost dist]$ kubectl exec -it test.rbd.csi.ceph.com-nodeplugin-dn4ww  -nceph-csi-operator-system -c log-rotator sh
kubectl exec [POD] [COMMAND] is DEPRECATED and will be removed in a future version. Use kubectl exec [POD] -- [COMMAND] instead.
sh-4.4# ls     
bin  csi-logs  dev  etc  home  lib  lib64  logrotate-config  lost+found  media	mnt  opt  proc	root  run  sbin  srv  sys  tmp	usr  var
sh-4.4# cd logrotate-config/
sh-4.4# cat csi 
/csi-logs/*.log {
	missingok
	compress
	copytruncate
	notifempty
	rotate 5
	daily
	maxsize 500M
}
sh-4.4# cd ..
sh-4.4# cd csi-logs/                  
sh-4.4# ls
csi-rbdplugin.log
sh-4.4# cat csi-rbdplugin.log 
Log file created at: 2024/08/16 08:10:18
Running on machine: minikube-m03
Binary: Built with gc go1.21.5 for linux/amd64
Log line format: [IWEF]mmdd hh:mm:ss.uuuuuu threadid file:line] msg
W0816 08:10:18.759665   17582 util.go:276] kernel 5.10.57 does not support required features
W0816 08:10:18.760132   17582 rbd_attach.go:252] kernel version "5.10.57" doesn't support cookie feature
sh-4.4# 
sh-4.4# 
sh-4.4# truncate -s 600M csi-rbdplugin.log 
sh-4.4# 
sh-4.4# ls -lh
total 4.0K
-rw-r--r-- 1 root root 600M Aug 16 08:15 csi-rbdplugin.log
sh-4.4# 
sh-4.4# exit
exit
[rider@localhost dist]$ kubectl logs  test.rbd.csi.ceph.com-ctrlplugin-54b744ff6c-chht6  -nceph-csi-operator-system -c log-rotator 
reading config file /logrotate-config/csi
Reading state from file: /var/lib/logrotate/logrotate.status
Allocating hash table for state file, size 64 entries

Handling 1 logs

rotating pattern: /csi-logs/*.log  after 1 days (5 rotations)
empty log files are not rotated, log files >= 524288000 are rotated earlier, old logs are removed
considering log /csi-logs/*.log
  log /csi-logs/*.log does not exist -- skipping
Creating new state
[rider@localhost dist]$ 
[rider@localhost dist]$ 
[rider@localhost dist]$ cd ..
[rider@localhost ceph-csi-operator]$ git add .
[rider@localhost ceph-csi-operator]$ 
[rider@localhost ceph-csi-operator]$ 
[rider@localhost ceph-csi-operator]$ 
[rider@localhost ceph-csi-operator]$ 
[rider@localhost ceph-csi-operator]$ kubectl exec -it test.rbd.csi.ceph.com-nodeplugin-dn4ww  -nceph-csi-operator-system -c log-rotator sh
kubectl exec [POD] [COMMAND] is DEPRECATED and will be removed in a future version. Use kubectl exec [POD] -- [COMMAND] instead.
sh-4.4# cd /csi-logs/
sh-4.4# ls
csi-rbdplugin.log  csi-rbdplugin.log.1.gz
sh-4.4# 


```

Madhu-1 (dismissed):

Madhu-1:
@nb-ohad PTAL and merge it if everything is good.

Madhu-1:
@parth-gr may i know the recent change you pushed after i approved the PR?

internal/controller/driver_controller.go
```diff
@@ -454,6 +512,17 @@ func (r *driverReconcile) reconcileControllerPluginDeployment() error {
...
+		logRotationEnabled := logRotationSpec != nil
+		controllerPluginPrivileged := pluginSpec.Privileged != nil
+		securityContext := utils.If(logRotationEnabled && controllerPluginPrivileged,
+			&corev1.SecurityContext{
+				Privileged: ptr.To(true),
+				Capabilities: &corev1.Capabilities{
+					Drop: []corev1.Capability{"All"},
+				},
+			},
+			nil)
```
nb-ohad, line 525: There are a couple of issues here:
1.  controllerPluginPrivileged but logRotate is not enabled you still need to set up the security context. 
2. Same for the opposite.
3. the `(` and `)` are not in the right place

internal/controller/driver_controller.go
```diff
@@ -741,7 +831,22 @@ func (r *driverReconcile) reconcileControllerPluginDeployment() error {
...
+						// CSI LogRotate Container
+						if logRotationEnabled {
+							resources := ptr.Deref(pluginSpec.Resources.LogRotator, corev1.ResourceRequirements{})
+							containers = append(containers, corev1.Container{
+								Name:            "log-rotator",
+								Image:           r.images["plugin"],
+								ImagePullPolicy: imagePullPolicy,
+								Resources:       resources,
+								Command:         []string{"/bin/bash", "-c"},
+								Args:            []string{`while true; do logrotate --verbose /logrotate-config/csi; sleep 15m; done`},
```
nb-ohad, line 843: There is no need for command and arg. Just have one command and move it into a const to be reused.

internal/controller/driver_controller.go
```diff
@@ -1026,6 +1161,22 @@ func (r *driverReconcile) reconcileNodePluginDeamonSet() error {
...
```
nb-ohad, line 1173: Same as the last comment

nb-ohad (changes requested):

internal/controller/driver_controller.go
parth-gr: >the ( and ) are not in the right place

Right formatting or right order?

internal/controller/driver_controller.go
```diff
@@ -454,6 +514,17 @@ func (r *driverReconcile) reconcileControllerPluginDeployment() error {
...
+		logRotationEnabled := logRotationSpec != nil
+		controllerPluginPrivileged := pluginSpec.Privileged != nil
+		securityContext := utils.If(
+			controllerPluginPrivileged,
+			&corev1.SecurityContext{
+				Privileged: ptr.To(true),
+				Capabilities: &corev1.Capabilities{
+					Drop: []corev1.Capability{"All"},
+				},
+			}, nil)
```
nb-ohad, line 527: We either need everything to have every container align with the `Privileged` flags which is not this change.
Or have containers that uses log rotate be privileged.
For now I will go with the other option:

```suggestion
		logRotateSecurityContext := utils.If(
			pluginSpec.Privileged != nil && controllerPluginPrivileged
			&corev1.SecurityContext{
				Privileged: ptr.To(true),
				Capabilities: &corev1.Capabilities{
					Drop: []corev1.Capability{"All"},
				},
			}, 
			nil,
		)
```

nb-ohad (changes requested):

internal/controller/driver_controller.go
parth-gr: DOne

internal/controller/driver_controller.go
```diff
@@ -454,6 +514,18 @@ func (r *driverReconcile) reconcileControllerPluginDeployment() error {
 		logVerbosity := ptr.Deref(r.driver.Spec.Log, csiv1a1.LogSpec{}).Verbosity
 		forceKernelClient := r.isCephFsDriver() && r.driver.Spec.CephFsClientType == csiv1a1.KernelCephFsClient
 		snPolicy := cmp.Or(r.driver.Spec.SnapshotPolicy, csiv1a1.VolumeSnapshotSnapshotPolicy)
+		logRotationSpec := cmp.Or(r.driver.Spec.Log, &csiv1a1.LogSpec{}).Rotation
+		logRotationEnabled := logRotationSpec != nil
+		logRotateSecurityContext := utils.If(
+			pluginSpec.Privileged != nil && logRotationEnabled,
+			&corev1.SecurityContext{
+				Privileged: ptr.To(true),
```
Madhu-1, line 522: what happens if the user sets the `.Privileged` to false?

parth-gr: I think we missed that condition we should check it explicitly.

internal/controller/driver_controller.go
```diff
@@ -72,6 +72,8 @@ const (
 // Annotation name for ownerref information
 const ownerRefAnnotationKey = "csi.ceph.io/ownerref"
 
+const logRotateCmd = `while true; do logrotate --verbose /logrotate-config/csi; sleep 15m; done`
```
nb-ohad, line 75: Please merge this const with the const above it into a single const section.

internal/controller/driver_controller.go
nb-ohad: @Madhu-1 If it is false, should we still have capabilities drop all?

Madhu-1: yes, that is correct, both are not dependent on each other.

parth-gr: Do we want this condition, 

`pluginSpec.Privileged != nil && *pluginSpec.Privileged`
OR
`pluginSpec.Privileged != nil && *pluginSpec.Privileged && logRotationEnabled`

For Drop all?

internal/controller/driver_controller.go
```diff
@@ -454,6 +515,18 @@ func (r *driverReconcile) reconcileControllerPluginDeployment() error {
...
+		logRotateSecurityContext := utils.If(
+			pluginSpec.Privileged != nil && *pluginSpec.Privileged && logRotationEnabled,
+			&corev1.SecurityContext{
+				Privileged: ptr.To(true),
+				Capabilities: &corev1.Capabilities{
+					Drop: []corev1.Capability{"All"},
+				},
+			},
+			nil,
+		)
```
Madhu-1, line 529: why not just do below

```go
pluginSecurityContext := &corev1.SecurityContext{
Privileged: ptr.To(false),
Capabilities: &corev1.Capabilities{
	Drop: []corev1.Capability{"All"},
	},
}

if pluginSpec.Privileged != nil && *pluginSpec.Privileged{
pluginSecurityContext=ptr.To(pluginSpec.Privileged)
}

```

internal/controller/driver_controller.go
```diff
@@ -1026,6 +1164,21 @@ func (r *driverReconcile) reconcileNodePluginDeamonSet() error {
 								})
 							}
 						}
+						// CSI LogRotate Container
+						if logRotationEnabled {
+							resources := ptr.Deref(pluginSpec.Resources.LogRotator, corev1.ResourceRequirements{})
+							containers = append(containers, corev1.Container{
+								Name:            "log-rotator",
+								Image:           r.images["plugin"],
+								ImagePullPolicy: imagePullPolicy,
```
Madhu-1, line 1173: set securityContext for this container as well like we are doing for other containers in the ds.

internal/controller/driver_controller.go
Madhu-1: Will revisit this later to set the SecurityContext for all the containers

internal/controller/driver_controller.go
Madhu-1: I will revisit this later as its internal chance, Its good to run container with least access.

Madhu-1 (dismissed):

internal/controller/driver_controller.go
nb-ohad, line 529:
```suggestion
		logRotateSecurityContext := utils.If(
			pluginSpec.Privileged != nil && *pluginSpec.Privileged && logRotationEnabled,
			&corev1.SecurityContext{
				Privileged: pluginSpec.Privileged,
				Capabilities: &corev1.Capabilities{
					Drop: []corev1.Capability{"All"},
				},
			},
			nil,
		)
```

nb-ohad:
LGTM
@Madhu-1 Are you ok with the state of the PR?

Madhu-1 (approved):

nb-ohad (approved):

PR #59, opened 2024-07-26
ci: update action to check git diff

Madhu-1:
updating github action to run git diff to check uncommitted files. The PR author sometimes forgot to push all the changes required as part of the PR. this causes inconsistency problems with the repo and local. This PR adds a check in the CI and also provides the `make` target to that developers can ensure that they are pushing all the required changes into the PR

### DIFF
--- a/.github/workflows/build.yaml
+++ b/.github/workflows/build.yaml
@@ -26,3 +26,8 @@ jobs:
           go-version-file: go.mod
       - name: run make build
         run: make build
+      - name: Check for uncommitted changes
+        run: make check-all-committed
+      - name: Show the uncommitted "git diff"
+        if: ${{ failure() }}
+        run: git diff ; false
--- a/Makefile
+++ b/Makefile
@@ -184,6 +184,10 @@ envtest: $(ENVTEST) ## Download setup-envtest locally if necessary.
 $(ENVTEST): $(LOCALBIN)
 	$(call go-install-tool,$(ENVTEST),sigs.k8s.io/controller-runtime/tools/setup-envtest,$(ENVTEST_VERSION))
 
+.PHONY: check-all-committed
+check-all-committed: ## Fail in case there are uncommitted changes
+	test -z "$(shell git status --short)" || (echo "files were modified: " ; git status --short ; false)
+
 .PHONY: golangci-lint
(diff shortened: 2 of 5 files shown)


nb-ohad:
@Madhu-1 What is the intention behind this action? The current description doesn't seem to provide any information about the problem you are trying to solve.
 
Where and how would uncommitted changes get into a CI run? Is that supposed to run locally? on Github? What am I missing?

Also, why do we have changes to openAPI CRD definition files in this PR? Is that on purpose? and mistake? 
If this PR is trying to accomplish two unrelated goals we should separate the content into two different PRs

Madhu-1:
> @Madhu-1 What is the intention behind this action? The current description doesn't seem to provide any information about the problem you are trying to solve.
> 
> Where and how would uncommitted changes get into a CI run? Is that supposed to run locally? on Github? What am I missing?
> 
> Also, why do we have changes to openAPI CRD definition files in this PR? Is that on purpose? and mistake? If this PR is trying to accomplish two unrelated goals we should separate the content into two different PRs

@nb-ohad after merging the last PR some changes are getting generated locally after running `make build` This is not expected, This is a check to ensure that the PR contains all the changes that are required not no changes left behind. For example check this action result https://github.com/ceph/ceph-csi-operator/actions/runs/10109935774/job/27958828177

nb-ohad:
> > @Madhu-1 What is the intention behind this action? The current description doesn't seem to provide any information about the problem you are trying to solve.
> > Where and how would uncommitted changes get into a CI run? Is that supposed to run locally? on Github? What am I missing?
> > Also, why do we have changes to openAPI CRD definition files in this PR? Is that on purpose? and mistake? If this PR is trying to accomplish two unrelated goals we should separate the content into two different PRs
> 
> @nb-ohad after merging the last PR some changes are getting generated locally after running `make build` This is not expected, This is a check to ensure that the PR contains all the changes that are required not no changes left behind. For example check this action result https://github.com/ceph/ceph-csi-operator/actions/runs/10109935774/job/27958828177

@Madhu-1 
I believe I understand now. This will block merge if the CI observes that running `make build` and/or `make lint-fix`, generates artifacts that are not parts of the code changes committed as part of the PR. 
That makes sense.

nb-ohad (approved):

PR #60, opened 2024-07-26
fix bugs found in controller

Madhu-1:
fix bugs found in nodeplugin, where the kubelet path was not specified and also the full driver name need to be present in the kubelet path for the kubelet to use cephcsi socket

### DIFF
--- a/internal/controller/driver_controller.go
+++ b/internal/controller/driver_controller.go
@@ -73,7 +73,7 @@ const ownerRefAnnotationKey = "csi.ceph.io/ownerref"
 
 // A regexp used to parse driver's prefix and type from the full name
 var nameRegExp, _ = regexp.Compile(fmt.Sprintf(
-	`^(?:(.+)\.)?(%s|%s|%s)\.csi\.ceph\.com$`,
+	`^(?:.+\.)?(%s|%s|%s)\.csi\.ceph\.com$`,
 	RbdDriverType,
 	CephFsDriverType,
 	NfsDriverType,
@@ -89,12 +89,11 @@ type DriverReconciler struct {
 type driverReconcile struct {
 	DriverReconciler
 
-	ctx          context.Context
-	log          logr.Logger
-	driver       csiv1a1.Driver
-	driverPrefix string
-	driverType   DriverType
-	images       map[string]string
+	ctx        context.Context
+	log        logr.Logger
+	driver     csiv1a1.Driver
+	driverType DriverType
+	images     map[string]string
 }
 
 // SetupWithManager sets up the controller with the Manager.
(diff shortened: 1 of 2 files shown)


leelavg:
the fix worked for me, however I suggest to add what were the issues.

internal/controller/driver_controller.go
```diff
@@ -963,10 +962,10 @@ func (r *driverReconcile) reconcileNodePluginDeamonSet() error {
 							utils.HostRunMountVolume,
 							utils.LibModulesVolume,
 							utils.KeysTmpDirVolume,
-							utils.PluginDirVolume(pluginSpec.KubeletDirPath, r.driverPrefix),
-							utils.PluginMountDirVolume(pluginSpec.KubeletDirPath),
-							utils.PodsMountDirVolume(pluginSpec.KubeletDirPath),
-							utils.RegistrationDirVolume(pluginSpec.KubeletDirPath),
+							utils.PluginDirVolume(kubeletDirPath, r.driver.Name),
```
nb-ohad, line 965: Hi @Madhu-1 
I am not sure about this. This is not the approach Rook takes and I would like to understand why we need the full name as part of the dir path? 

If the issue you are trying to solve is that it is missing for the drivers without a prefix we can fix that.

Madhu-1: @nb-ohad can you please point me to the code in Rook where do you see that we are using the prefix not the driver's full name in the volume?

nb-ohad: Ok, I was wrong. Rook does use the prefix but then in the template it concat the entire suffix as well 
image

nb-ohad: @Madhu-1 As the change completely removes the prefix usage from the codebase, can you please remove the capture from the regexp? There is no need to have it in the match list if we will not use it.

Madhu-1: Done.

internal/controller/driver_controller.go
```diff
@@ -73,7 +73,7 @@ const ownerRefAnnotationKey = "csi.ceph.io/ownerref"
 
 // A regexp used to parse driver's prefix and type from the full name
 var nameRegExp, _ = regexp.Compile(fmt.Sprintf(
-	`^(?:(.+)\.)?(%s|%s|%s)\.csi\.ceph\.com$`,
+	`(%s|%s|%s)\.csi\.ceph\.com$`,
```
nb-ohad, line 76: This makes the regexp too permissive. It will allow names like 'mydriverrbd.csi.ceph.com`.
It is also preferable, for performance reasons, to compile a regexp that matches the entire name and not just the suffix. This way we eliminate a class of bugz that relates to greedy grepping 

```suggestion
	`^(?:.+\.)?(%s|%s|%s)\.csi\.ceph\.com$`,
```

nb-ohad (changes requested):

nb-ohad (approved):

PR #68, opened 2024-07-30
export default SecurityContextConstraints for OCP clusters

weirdwiz:

This PR exports the SecurityContextConstraints for admins deploying the ceph-csi-operators in OCP clusters

Do you have any questions?

For example:

## Related issues ##

Fixes: #issue_number

## Future concerns ##

**Checklist:**

  guidelines in the [developer
* [ ] [Pending release


api/go.mod
```diff
@@ -0,0 +1,27 @@
+module github.com/ceph/ceph-csi-operator/api
+
+go 1.22.5
```
Madhu-1, line 3: have like this https://github.com/ceph/ceph-csi-operator/blob/1dda58616ce3c72b7903b439267ca3e46e1dbd0a/go.mod#L3-L4

api/ocp/scc.go
```diff
@@ -0,0 +1,62 @@
...
+	"fmt"
+
+	secv1 "github.com/openshift/api/security/v1"
+	corev1 "k8s.io/api/core/v1"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+)
+
+// NewSecurityContextConstraints exports a default configuration of the
+// Security Context Constraints to deploy ceph-csi-operator in an OCP cluster
+func NewSecurityContextConstraints(namespace string, name string) *secv1.SecurityContextConstraints {
```
Madhu-1, line 13: we need to scc.yaml in config folder which is required for upstream users and we can use it as template here and make it customizable with serviceaccount name and namespace.

Madhu-1 (changes requested):

api/go.mod
nb-ohad, line 3: Why the bump from 1.22.0 to 1.22.5?
And if we do want to do it, lets have a separate PR for it

api/ocp/scc.go
nb-ohad, line 1: We never discussed having code to generate an SCC, we discussed having a static yaml file with the content. 
@weirdwiz Please offer justification for the ask to merge this contribution.

nb-ohad (changes requested):

api/go.mod
weirdwiz: updated with this change

api/ocp/scc.go
weirdwiz: add scc.yaml as a template, initially followed that route. but importing it threw some errors, so pivoted to generating the yaml from go code. 
but there was some issue with me pushing the package, that resulted the issue. pivoted back to the template

api/go.mod
weirdwiz: fixed, moved it back. and set it up just like the parent package

api/ocp/scc.go
Madhu-1: There are some improvement required, lets talk about it in today's standup

api/ocp/scc.yaml
```diff
@@ -0,0 +1,37 @@
...
+seLinuxContext:
+  type: RunAsAny
+supplementalGroups:
+  type: RunAsAny
+users:
+  - system:serviceaccount:{{.Namespace}}:csi-rbd-ctrlplugin-sa
+  - system:serviceaccount:{{.Namespace}}:csi-cephfs-ctrlplugin-sa
+  - system:serviceaccount:{{.Namespace}}:csi-nfs-ctrlplugin-sa
+  - system:serviceaccount:{{.Namespace}}:csi-rbd-nodeplugin-sa
+  - system:serviceaccount:{{.Namespace}}:csi-cephfs-nodeplugin-sa
```
Madhu-1, line 31: Having a template here doesnt help upstream Rook users. we need complete yaml where Rook or user can use it directly

nb-ohad: +1

nb-ohad:
@weirdwiz This PR is trying to attack multiple things at once:
- Make the API folder its own package
- Add SCC template

Please separate into two different PRs

api/ocp/scc.go
```diff
@@ -0,0 +1,37 @@
...
+	}
+
+	scc := &secv1.SecurityContextConstraints{}
+	err = yaml.UnmarshalStrict(buf.Bytes(), scc)
+	if err != nil {
+		return nil, fmt.Errorf("error unmarshaling YAML: %v", err)
+	}
+
+	return scc, nil
+}
```
nb-ohad, line 37: Why do we need this? where is it used? 
The assumption is that a template is enough for users. 
They will either add it to their own automation or use it as part of a halm chart

nb-ohad (changes requested):

api/ocp/scc.go
weirdwiz: the ocs-operator imports all the SCCs and uses this function, to apply it during initialization

ref: https://github.com/red-hat-storage/ocs-operator/blob/main/controllers/ocsinitialization/sccs.go

api/ocp/scc.go
```diff
@@ -0,0 +1,36 @@
...
+	secv1 "github.com/openshift/api/security/v1"
+	"sigs.k8s.io/yaml"
+)
+
+//go:embed scc.yaml
+var sccYAMLTemplate string
+
+// NewSecurityContextConstraints loads the embedded SCC YAML template, replaces the namespace,
+// and returns it as a SecurityContextConstraints object
+func NewSecurityContextConstraints(name string, namespace string) (*secv1.SecurityContextConstraints, error) {
```
Madhu-1, line 17: what i was expecting was like below
```go
//go:embed scc.yaml
var sccYAMLTemplate string

// GetSecurityContextConstraintsTemplate returns the embedded SCC YAML template as a string.
func GetSecurityContextConstraintsTemplate() string {
	return sccYAMLTemplate
}
```

and let the user import it and unmarshal and do generate the below details

```
- system:serviceaccount:ceph-csi-operator-system:csi-rbd-ctrlplugin-sa
  - system:serviceaccount:ceph-csi-operator-system:csi-cephfs-ctrlplugin-sa
  - system:serviceaccount:ceph-csi-operator-system:csi-nfs-ctrlplugin-sa
  - system:serviceaccount:ceph-csi-operator-system:csi-rbd-nodeplugin-sa
  - system:serviceaccount:ceph-csi-operator-system:csi-cephfs-nodeplugin-sa
  - system:serviceaccount:ceph-csi-operator-system:csi-nfs-nodeplugin-sa
```

The service account and namespace and service account name is what others need to worry about.

@nb-ohad Is this what you also have in mind? @leelavg Does this works well for you?

nb-ohad: @Madhu-1 The question here is: why should anyone have to add a code dependency on this project to consume one static yaml template file? 

My alternative is. have the file in this project. Instruct other projects to pull this file into their project as part of their build phase using something like `curl get` directly from github (or something similar)  

Regarding the code itself. I don't believe the above is going to work. The second code base is going to pull this code as a dependency but not the temple file (As it is not a go file). This means that the embed command will fail and the project will not build. 

Unfortunately, if you want to provide it via code, we will need to embed the entire content in a multi-line const string.

nb-ohad:
@weirdwiz Can you please rebase this and get rid of the first commit? (which is already addressed on a different PR that got merged)

Madhu-1:
@weirdwiz ping

PR #69, opened 2024-07-30
Driver Controller: Fix support for SnapshotPolicy settings

nb-ohad:

The existing code ignores the driver's SnapshotPolicy settings. This PR fix the behavior.

No

### DIFF
--- a/api/v1alpha1/driver_types.go
+++ b/api/v1alpha1/driver_types.go
@@ -72,10 +72,6 @@ const (
 	// Disables the feature and remove the snapshotter sidercar
 	NoneSnapshotPolicy SnapshotPolicyType = "none"
 
-	// Inspect the CRD's for volumesnapshot and volumegroupsnapshot and enable
-	// corresponding features (will results in deployment of a snapshotter sidecar)
-	AutoDetectSnapshotPolicy SnapshotPolicyType = "autodetect"
-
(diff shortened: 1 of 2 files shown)


nb-ohad:
@Madhu-1 I have selected `VolumeSnapshot` snapshot policy as the default value. If you feel `None` is a better choice please reply here and I will make the change.

api/v1alpha1/driver_types.go
```diff
@@ -72,10 +72,6 @@ const (
 	// Disables the feature and remove the snapshotter sidercar
 	NoneSnapshotPolicy SnapshotPolicyType = "none"
 
-	// Inspect the CRD's for volumesnapshot and volumegroupsnapshot and enable
-	// corresponding features (will results in deployment of a snapshotter sidecar)
-	AutoDetectSnapshotPolicy SnapshotPolicyType = "autodetect"
```
Madhu-1, line 77: @nb-ohad now is it responsibility of the user to check if the CRDs exist and set the policy?

nb-ohad: Yes, but we might want to re-introduce this in the future.

The reason I dropped it for now is that it is more complicated than just looking for the CRD existence. We also need to make sure the images you are using support the feature, and report your detection on the status, maybe even with a reason. If we want the entire auto-detect mechinizem will be "magic" from a user perspective.

Madhu-1 (approved):

PR #72, opened 2024-08-01
Use list of BlockPoolMapping in clientProfileMappings spec

sp98:
Changes BlockPoolMappings spec to list in the API

Do you have any questions?

For example:

## Related issues ##

Fixes: #issue_number

## Future concerns ##

**Checklist:**

  guidelines in the [developer
* [ ] [Pending release

### DIFF
--- a/api/v1alpha1/clientprofilemapping_types.go
+++ b/api/v1alpha1/clientprofilemapping_types.go
@@ -42,7 +42,7 @@ type BlockPoolMappingSpec struct {
 // ClientProfileMappingSpec defines the desired state of ClientProfileMapping
 type ClientProfileMappingSpec struct {
 	//+kubebuilder:validation:Optional
-	BlockPoolMapping *BlockPoolMappingSpec `json:"blockPoolMapping,omitempty"`
+	BlockPoolMapping []BlockPoolMappingSpec `json:"blockPoolMapping,omitempty"`
 }
 
 // ClientProfileMappingStatus defines the observed state of ClientProfileMapping
--- a/api/v1alpha1/zz_generated.deepcopy.go
+++ b/api/v1alpha1/zz_generated.deepcopy.go
@@ -305,8 +305,8 @@ func (in *ClientProfileMappingSpec) DeepCopyInto(out *ClientProfileMappingSpec) 
 	*out = *in
 	if in.BlockPoolMapping != nil {
 		in, out := &in.BlockPoolMapping, &out.BlockPoolMapping
-		*out = new(BlockPoolMappingSpec)
-		**out = **in
+		*out = make([]BlockPoolMappingSpec, len(*in))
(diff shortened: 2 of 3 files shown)


api/v1alpha1/clientprofilemapping_types.go
```diff
@@ -42,7 +42,7 @@ type BlockPoolMappingSpec struct {
 // ClientProfileMappingSpec defines the desired state of ClientProfileMapping
 type ClientProfileMappingSpec struct {
 	//+kubebuilder:validation:Optional
-	BlockPoolMapping *BlockPoolMappingSpec `json:"blockPoolMapping,omitempty"`
+	BlockPoolMapping *[]BlockPoolMappingSpec `json:"blockPoolMapping,omitempty"`
```
Madhu-1, line 45: @sp98 why did the user cant create multiple CR's instead of this change?

sp98: Single CR for all the pool IDs looks like a much easier option to manage. Having multiple CRs for each pool ID mappings would be difficult to manage. But it also depends on how csi-operator wants to handle it.

sp98: Currently we store the mappings in the configmap like this: 
```
 oc get cm rook-ceph-csi-mapping-config   -o yaml
apiVersion: v1
data:
  csi-mapping-config-json: '[{"ClusterIDMapping":{"openshift-storage":"openshift-storage"},"RBDPoolIDMapping":[{"1":"1"}]}]'
kind: ConfigMap
metadata:
```

The the equivalent `clientProfileMappings` CR can look like this: 
```
kind: ClientProfileMapping
apiVersion: csi.ceph.io/v1
metadata:
  name: storage
  namespace: <operator-namespace>
spec:
  blockPoolMapping:
  - local:
      clientPorfileName: openshift-storage
      poolID: 1
    remote:
      clientProfileName: openshift-storage
      poolID: 1
``` 

And if more pool are created, then we add the mappings in the same CR like below: 
```
kind: ClientProfileMapping
apiVersion: csi.ceph.io/v1
metadata:
  name: storage
  namespace: <operator-namespace>
spec:
  blockPoolMapping:
  - local:
      clientPorfileName: openshift-storage
      poolID: 1
    remote:
      clientProfileName: openshift-storage
      poolID: 1
 - local:
      clientPorfileName: openshift-storage
      poolID: 2
    remote:
      clientProfileName: openshift-storage
      poolID: 3
```

sp98: make sense.

Madhu-1: After thinking a little more on this one i see both pros and cons, having a list makes sense but the admin needs to decide on what they want. Lets discuss in today's stand up

PR #73, opened 2024-08-01
Update ci test

Madhu-1:

This updates the CI job to build and deploy the cephcsi operator and check if it's running or not.

### DIFF
--- a/test/scripts/utils.sh
+++ b/test/scripts/utils.sh
@@ -1,0 +1,90 @@
+#!/bin/bash
+
+KUBECTL_RETRY=5
+KUBECTL_RETRY_DELAY=10
+
+# kubectl_retry calls `kubectl` with the passed arguments. In case of a
+# failure, the `kubectl` command will be retried for `KUBECTL_RETRY` times,
+# with a delay of `KUBECTL_RETRY_DELAY` between them.
+#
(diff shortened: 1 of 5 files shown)


test/scripts/utils.sh
```diff
@@ -0,0 +1,90 @@
...
+        # empty stderr for the next loop
+        true > "${stderr}"
+        echo "$(date): 'kubectl_retry ${action} ${*}' failed (${retries}/${KUBECTL_RETRY}), will retry in ${KUBECTL_RETRY_DELAY} seconds" > /dev/stderr
+
+        sleep ${KUBECTL_RETRY_DELAY}
+
+        # reset ret so that a next working kubectl does not cause a non-zero
+        # return of the function
+        ret=0
+    done
```
nb-ohad, line 78: Shouldn't we look for the process (kubectl) exit value to determine failures instead of searching `std*`?

nb-ohad (changes requested):

test/scripts/utils.sh
nb-ohad: As this code is already proven to work in other repos we agreed offline to take it as is and look at enhancements in the future.

nb-ohad (approved):

Issue #80, opened 2024-08-02
Topology should not be enabled by default in provisioner


leelavg:
Hitting below when topology is enabled by default
```
E0802 11:27:17.599933       1 controller.go:957] error syncing claim "e284c95e-89b7-4ad8-9b7a-5c02970bd48e": failed to provision volume with StorageClass "ocs-storagecluster-ceph-rbd": error generating accessibility requirements: no available topology found
```

Statement from Madhu: `It should be set to false by default and enable it only if set domain label in CR for topology`

Madhu-1:
Sorry, i didn't notice this one but this and https://github.com/ceph/ceph-csi-operator/issues/83 are tracking the same.

PR #91, opened 2024-08-05
api: add support for topology based provisioning

Madhu-1:
This PR exposes new fields in the nodeplugin spec where the user can specify the domain labels for the cephcsi to advertise its topology domain

fixes: #80


api/v1alpha1/driver_types.go
```diff
@@ -166,6 +172,9 @@ type NodePluginSpec struct {
 	// Control the host mount of /etc/selinux for csi plugin pods. Defaults to false
 	//+kubebuilder:validation:Optional
 	EnableSeLinuxHostMount *bool `json:"EnableSeLinuxHostMount,omitempty"`
+	// Topology settings for the plugin pods
+	//+kubebuilder:validation:Optional
+	Topology *TopologySpec `json:"topology,omitempty"`
```
nb-ohad, line 177: Is topology only important for NodePlugin? what about the ControllerPlugin? shouldn't er have that in the PodCommonSpec?

internal/utils/csi.go
```diff
@@ -393,3 +392,11 @@ func FuseMountOptionsContainerArg(options map[string]string) string {
 		"",
 	)
 }
+
+func TopologyDomainLaels(options []string) string {
```
nb-ohad, line 396: 1. a misspelling of the word "labels"
2. We should align on the naming convention for dynamic container args (`<arg name>+ContainerArg`) 
```suggestion
func DomainLabelsContainerArg(options []string) string {
```

nb-ohad (changes requested):

api/v1alpha1/driver_types.go
Madhu-1: This is only required for node plugin they need to advertise topology on the node where it is running

internal/utils/csi.go
```diff
@@ -393,3 +392,11 @@ func FuseMountOptionsContainerArg(options map[string]string) string {
 		"",
 	)
 }
+
```
nb-ohad, line 395:
```suggestion
```

api/v1alpha1/driver_types.go
nb-ohad: ack

nb-ohad (approved):

PR #93, opened 2024-08-05
fix: address csi addons problems in csi

Madhu-1:
fix different problems found during testing for the csiaddons.

PS: As it is small fixes not creating different commits


internal/controller/driver_controller.go
```diff
@@ -434,6 +434,14 @@ func (r *driverReconcile) reconcileControllerPluginDeployment() error {
 			utils.LeaderElectionRetryPeriodContainerArg(leaderElectionSpec.RetryPeriod),
 		}
 
+		// csiaddons does not require utils.LeaderElectionContainerArg as its enabled by default and no option to disable it
+		csiAddonsLeaderElection := []string{
+			utils.LeaderElectionNamespaceContainerArg(r.driver.Namespace),
+			utils.LeaderElectionLeaseDurationContainerArg(leaderElectionSpec.LeaseDuration),
+			utils.LeaderElectionRenewDeadlineContainerArg(leaderElectionSpec.RenewDeadline),
+			utils.LeaderElectionRetryPeriodContainerArg(leaderElectionSpec.RetryPeriod),
+		}
```
nb-ohad, line 443: This seems like a duplication of the `leaderElectionArg` setting. What is the difference?
Also, if there is a difference I am missing, please embed these settings in the csiAddons container arg instead of having it here as it only applies to that container.

internal/controller/driver_controller.go
```diff
@@ -613,9 +621,9 @@ func (r *driverReconcile) reconcileControllerPluginDeployment() error {
 								Image:           r.images["addons"],
 								ImagePullPolicy: imagePullPolicy,
 								Args: append(
-									slices.Clone(leaderElectionArgs),
+									slices.Clone(csiAddonsLeaderElection),
```
nb-ohad, line 624: if these args are used only once, here, there is no need to define it on top and clone it here. Just have it directly here in the first place

nb-ohad (changes requested):

internal/controller/driver_controller.go
Madhu-1: There was a comment added what we have different from the leaderElectionArg, i think you missed that

nb-ohad: I did, thank you

nb-ohad (approved):

PR #94, opened 2024-08-05
api: make apis into a seperate package

weirdwiz:
# API Types Moved to Separate Package

Moves API types to a new `api` package, improving project structure and separation of concerns.

Key changes:
- New `api` directory with own `go.mod`
- Updated import paths
- Added `replace` directive in root `go.mod`

# Context
Aligns with Kubernetes operator development practices. Facilitates independent versioning of API types.

# Special Attention
Review updated import paths and new `replace` directive.

# Backward Compatibility
Compatible. External projects may need import path updates.

# External Context
Follows common practice in Kubernetes ecosystem for API separation.

# Future Concerns
- None

**Checklist:**
- [x] Commit messages follow guidelines
- [x] Developer guide on PRs reviewed
- [ ] Release notes updated (if needed)
- [ ] Documentation updated (if needed)
- [ ] Tests added/updated (if needed)


api/go.mod
```diff
@@ -0,0 +1,26 @@
+module github.com/ceph/ceph-csi-operator/api
+
+go 1.22.5
```
Madhu-1, line 3: can you pleae use https://github.com/ceph/ceph-csi-operator/blob/4919ece1961894cca23a1bf13d80aff1bac58304/go.mod#L3-L4 same here

api/go.mod
```diff
@@ -0,0 +1,28 @@
+module github.com/ceph/ceph-csi-operator/api
```
Madhu-1, line 1: i think you need to update the go.mod in the root folder to use this as a replace section. can you please check

nb-ohad:
@weirdwiz Please fill in the relevant information in the PR description. Right now the PR description is just the empty PR template.

api/go.mod
```diff
@@ -0,0 +1,28 @@
...
+
+require (
+	k8s.io/api v0.30.3
+	k8s.io/apimachinery v0.30.3
+	sigs.k8s.io/controller-runtime v0.18.4
+)
+
+require (
+	github.com/go-logr/logr v1.4.1 // indirect
+	github.com/gogo/protobuf v1.3.2 // indirect
```
nb-ohad, line 15: You should not have 2 required sections inside a mode file. Please merge both into 1

nb-ohad (changes requested):
Why do we need a `vendor` folder in the API package? can't this new package be vendorless?

api/go.mod
weirdwiz: this is auto generated, if i run go mod tidy
the require gets split again

api/go.mod
weirdwiz: updated the PR, closing this conversation

api/go.mod
Madhu-1: There we will 2 require one for direct dependencies and another for indirect dependencies

Madhu-1 (approved):

api/go.mod
nb-ohad: Ack, thanks for the clarification

weirdwiz:
@Madhu-1 updated the dockerfile

Madhu-1 (approved):

nb-ohad (approved):

leelavg:
All, we have put an extra effort in ocs-operator APIs not to depend on controller-runtime I think it was worth it as controller-runtime frequently (when compared to APIs) introduces breaking changes.

I remember projects backing off imports due to controller-runtime rather than API/more vendor.

Issue #89, opened 2024-08-05
LivnessProbe fields and grpcTimeout vaules should be in time.Duration unit


subhamkrai:
Any reason `LivenessProbe` https://github.com/ceph/ceph-csi-operator/blob/204ddfa4cf8af796ff70ec5d5014a5cad1cb4d14/api/v1alpha1/driver_types.go#L228-L245 and `grpcTimeout` https://github.com/ceph/ceph-csi-operator/blob/204ddfa4cf8af796ff70ec5d5014a5cad1cb4d14/api/v1alpha1/driver_types.go#L284  in not `time. Duration`, when rook is setting default values which of ` time.Duration` it is failing

rook error logs from csi-pod
```
 kc logs  rook-ceph.cephfs.csi.ceph.com-ctrlplugin-6ffb4b774-86vkd -c csi-resizer
invalid value "137000000000s" for flag -leader-election-lease-duration: parse error
```

and

```
kc logs rook-ceph.rbd.csi.ceph.com-ctrlplugin-5744b6bb4-vs87q -c csi-resizer
invalid value "150000000000s" for flag -timeout: parse error
Usage of /csi-resizer:
```

Issue #95, opened 2024-08-06
SnapshotPolicyType Const  value should match kubebuilder enum


subhamkrai:
The defined `SnapshotPolicyType` is https://github.com/ceph/ceph-csi-operator/blob/4919ece1961894cca23a1bf13d80aff1bac58304/api/v1alpha1/driver_types.go#L72-L79 but kubebuilder enum has different rules with camelCase
https://github.com/ceph/ceph-csi-operator/blob/4919ece1961894cca23a1bf13d80aff1bac58304/api/v1alpha1/driver_types.go#L297-L298

Madhu-1:
@subhamkrai do you have a plan to fix this?

subhamkrai:
> @subhamkrai do you have a plan to fix this?

yes I will do

PR #97, opened 2024-08-06
api: fix camelCase issue SnapshotPolicyType

subhamkrai:

 there was cameCase type in SnapshotPolicyType enum validation.

no

For example:

## Related issues ##

Fixes: #89 #95 

## Future concerns ##

  guidelines in the [developer
* [ ] [Pending release

### DIFF
--- a/api/v1alpha1/driver_types.go
+++ b/api/v1alpha1/driver_types.go
@@ -294,7 +294,7 @@ type DriverSpec struct {
 
 	// Select a policy for snapshot behavior: none, autodetect, snapshot, sanpshotGroup
 	//+kubebuilder:validation:Optional
-	//+kubebuilder:validation:Enum:=none;autodetect;volumegroupsnapshot;volumesnapshot
+	//+kubebuilder:validation:Enum:=none;volumeGroupSnapshot;volumeSnapshot
 	SnapshotPolicy SnapshotPolicyType `json:"snapshotPolicy,omitempty"`
 
 	// OMAP generator will generate the omap mapping between the PV name and the RBD image.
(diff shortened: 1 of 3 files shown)


nb-ohad:
/hold
@subhamkrai Why are we making these changes? Why do we need the values to be lowercased and why use duration?

Madhu-1:
> /hold @subhamkrai Why are we making these changes? Why do we need the values to be lowercased and why use duration?

all the containers expect the timeout to be in time.Duration format and this is an extra restriction to do the conversion by self and set only the value in seconds here, if we have it in time.Duration its user-friendly as well.

nb-ohad:
As discussed offline, I am not objecting to the change as long as we pair it with validations to make sure the user must match a duration string structure.

api/v1alpha1/driver_types.go
```diff
@@ -294,7 +294,7 @@ type DriverSpec struct {
 
 	// Select a policy for snapshot behavior: none, autodetect, snapshot, sanpshotGroup
 	//+kubebuilder:validation:Optional
-	//+kubebuilder:validation:Enum:=none;autodetect;volumegroupsnapshot;volumesnapshot
+	//+kubebuilder:validation:Enum:=none;autodetect;volumeGroupSnapshot;volumeSnapshot
```
nb-ohad, line 297: @subhamkrai I believe `autodetect` needs to be removed from the validation as it is not a valid value anymore.
Can you please add this change to the PR?

nb-ohad (changes requested):

api/v1alpha1/driver_types.go
subhamkrai: done

Madhu-1 (approved):

nb-ohad (approved):

Madhu-1 (approved):

Issue #44, opened 2024-07-18
Add length check on the driver CR name


Madhu-1:
we need to have length check on the driver CR name to keep it minimal, we can do this as followup as well.

_Originally posted by @Madhu-1 in https://github.com/ceph/ceph-csi-operator/pull/24#discussion_r1682405782_

PR #98, opened 2024-08-06
add name length validation for driver

Madhu-1:

Choosing an arbitrary length which is 56 for the csi driver name and adding a validation to block CR creation i the name has more chars.

Do you have any questions?

No

Yes but only to an extent

Yes, the migration not be possible for the existing driver if names are beyond 63 chars, or else we don't have a problem.

## Related issues ##

Fixes: #44


Madhu-1:
Seeking more advice and it's a hard restriction with names and am open to another number as well. @nb-ohad @travisn @blaineexe

Madhu-1:
we create svc from the driver name which can be a maximum 63 chars, we also append `liveness` to it which is 8, with this restriction 54+8=62

nb-ohad:
/hold 
I believe this needs more discussion

BlaineEXE:
54 chars seems like plenty to me.

Madhu-1:
For me, this makes sense as well and not to think too much about it, if someone wants to have a different solution am going to close it and let others do it.

Madhu-1:
Closing it as am not working on it anymore

Issue #88, opened 2024-08-05
failed to create livness service


subhamkrai:
the operator is failing to create service
```
2024-08-05T09:53:09Z	ERROR	Failed to create livness service	{"controller": "driver", "controllerGroup": "csi.ceph.io", "controllerKind": "Driver", "Driver": {"name":"rook-ceph.cephfs.csi.ceph.com","namespace":"rook-ceph"}, "namespace": "rook-ceph", "name": "rook-ceph.cephfs.csi.ceph.com", "reconcileID": "4a4c7e95-d4e0-469e-8189-e58300b020a5", "service": "rook-ceph.cephfs.csi.ceph.com-liveness", "error": "Service \"rook-ceph.cephfs.csi.ceph.com-liveness\" is invalid: metadata.name: Invalid value: \"rook-ceph.cephfs.csi.ceph.com-liveness\": a DNS-1035 label must consist of lower case alphanumeric characters or '-', start with an alphabetic character, and end with an alphanumeric character (e.g. 'my-name',  or 'abc-123', regex used for validation is '[a-z]([-a-z0-9]*[a-z0-9])?')"}
github.com/ceph/ceph-csi-operator/internal/controller.logCreateOrUpdateResult
	/workspace/internal/controller/driver_controller.go:1098
github.com/ceph/ceph-csi-operator/internal/controller.(*driverReconcile).reconcileLivenessService
	/workspace/internal/controller/driver_controller.go:1037
github.com/ceph/ceph-csi-operator/internal/utils.RunConcurrently.func1
	/workspace/internal/utils/core.go:37
[...]

PR #99, opened 2024-08-06
generate service name as per kubernetes

Madhu-1:

kubernetes doesnt allow to have any special chars other than `-` in the names, we need to replace all special chars that we allow with `-` for service names.

## Related issues ##

Fixes: #88

### DIFF
--- a/internal/controller/driver_controller.go
+++ b/internal/controller/driver_controller.go
@@ -1083,7 +1083,7 @@ func (r *driverReconcile) reconcileNodePluginDeamonSet() error {
 func (r *driverReconcile) reconcileLivenessService() error {
 	service := &corev1.Service{}
 	service.Namespace = r.driver.Namespace
-	service.Name = r.generateName("liveness")
+	service.Name = r.generateServiceName("liveness")
 
 	log := r.log.WithValues("service", service.Name)
 	log.Info("Reconciling liveness service")
@@ -1136,6 +1136,17 @@ func (r *driverReconcile) isNfsDriver() bool {
 
 func (r *driverReconcile) generateName(suffix string) string {
 	return fmt.Sprintf("%s-%s", r.driver.Name, suffix)
+}
+
+// generateServiceName generates a service name by replacing all special characters with a hyphen
+// as required by Kubernetes for service name, This assumes that the name
+// from the generateName is RFC 1123 compliant.
+func (r *driverReconcile) generateServiceName(suffix string) string {
+	name := r.generateName(suffix)
+	// Define a regex pattern to match all special characters except hyphen
+	re := regexp.MustCompile(`[^a-z0-9-]`)
+	// Replace all special characters with a hyphen
+	return re.ReplaceAllString(name, "-")
 }
 
 func getControllerPluginPodAffinity(


internal/controller/driver_controller.go
```diff
@@ -1066,6 +1070,22 @@ func (r *driverReconcile) generateName(suffix string) string {
...
+	re, err := regexp.Compile(`[^\w\s]`)
+	if err != nil {
+		return "", fmt.Errorf("failed to compile regex: %v", err)
+	}
+
+	// Replace all special characters with a hyphen
+	result := re.ReplaceAllString(name, "-")
+
+	return result, nil
+}
```
nb-ohad, line 1087: If we use this only once, why do we need a helper function for it? 
If we do want a helper function why do we need it on the `driverReconcile` object?

internal/controller/driver_controller.go
```diff
@@ -1066,6 +1070,22 @@ func (r *driverReconcile) generateName(suffix string) string {
...
 
+// generateServiceName generates a service name by replacing all special characters with a hyphen
+// as required by Kubernetes for service name.
+func (r *driverReconcile) generateServiceName(suffix string) (string, error) {
+	name := r.generateName(suffix)
+	// Define a regex pattern to match all special characters
+	re, err := regexp.Compile(`[^\w\s]`)
+	if err != nil {
+		return "", fmt.Errorf("failed to compile regex: %v", err)
+	}
```
nb-ohad, line 1081: This seems wrong, you are selecting `\w` which in regexp selected all valid word chars

nb-ohad (changes requested):

internal/controller/driver_controller.go
Madhu-1: This is because of generateName which is dependent on driverReconcile and helper function help is to keep it simple and we can write unit tests for it. and also keep the main provisioner reconcile simple one as well.

internal/controller/driver_controller.go
Madhu-1: All the validation already done on the name, to keep it simple just replacing all the special chars with `-`

nb-ohad: Something here is still not clear to me. Your regexp selects both `\w` and `\s` and then negates the selection. 
How does that fit into the category of all special chars?

Madhu-1: Let me make it simple and remove `\s` which seems to cause problem and driver should take care of it and replaces `\w` with `[^a-zA-Z0-9-]` where `/w` was allowing `_` which is also not valid

Madhu-1:
> /hold working on different log to generate names for resources

 not require, we don't need to hold this as its required in any case

internal/controller/driver_controller.go
nb-ohad: @Madhu-1 
Checked the k8s docs. A service name must follow RFC 1035. This means the following:

- contain at most 63 characters
- contain only lowercase alphanumeric characters or '-'
- start with an alphabetic character
- end with an alphanumeric character

So lets add lowercasing of the name to your code

internal/controller/driver_controller.go
nb-ohad: Ok, so can we change the approach and have a function we send an existing name and the function align it with RFC 1.35.?

This way we can move from:
```func (r *driverReconcile) generateServiceName(suffix string) (string, error) {```
to 
```func  adjustNameforRfc1035(name) string {``` 

and move it to the util package

File: internal/controller/driver_controller.go [hunk 3]
```diff
@@ -1138,6 +1142,22 @@ func (r *driverReconcile) generateName(suffix string) string {
...
 }
 
+// generateServiceName generates a service name by replacing all special characters with a hyphen
+// as required by Kubernetes for service name.
+func (r *driverReconcile) generateServiceName(suffix string) (string, error) {
+	name := r.generateName(suffix)
+	// Define a regex pattern to match all special characters except hyphen
+	re, err := regexp.Compile(`[^a-z0-9-]`)
+	if err != nil {
+		return "", fmt.Errorf("failed to compile regex: %v", err)
```
nb-ohad, line 1152: If this fails it means that we coded it wrong (as the regexp is hard-coded)! not that we have a runtime error. 
In this case, a panic is preferable because it is surfaced to the developer faster and more aggressively. 

Can we change to a panic here and drop the error from the function?

nb-ohad (changes requested):

internal/controller/driver_controller.go
Madhu-1: IMHO this is not required we are trying to simply the thing too much

internal/controller/driver_controller.go
Madhu-1: panic causes the major problem and does not fix anything, I don't want to stop the whole operator running for this one, if a specific thing is failing let it fail and surface the error to the specific CR not cause problems for others, The panic will not help in all environments.

internal/controller/driver_controller.go
```diff
@@ -1138,6 +1142,22 @@ func (r *driverReconcile) generateName(suffix string) string {
 	return fmt.Sprintf("%s-%s", r.driver.Name, suffix)
 }
 
+// generateServiceName generates a service name by replacing all special characters with a hyphen
+// as required by Kubernetes for service name.
+func (r *driverReconcile) generateServiceName(suffix string) (string, error) {
+	name := r.generateName(suffix)
+	// Define a regex pattern to match all special characters except hyphen
+	re, err := regexp.Compile(`[^a-z0-9-]`)
```
nb-ohad, line 1150: This will replace uppercase letters with `-`. This is not the desired behavior. 
What we want is to allow uppercase letters but lowercase the end result after / before the substitution

nb-ohad (changes requested):

internal/controller/driver_controller.go
nb-ohad: That will never happen, panic will ensure that. If a developer makes a typing mistake (which is the only type of mistake that can happen here) panic guarantees it is observed. 

The alternative is that a typing mistake creates a logic mistake that can only be observed by test cases

internal/controller/driver_controller.go
Madhu-1: AFAIK upper case is not allowed in the SVC name, Do you have an example where SVC with upper case is allowed?

internal/controller/driver_controller.go
Madhu-1: yes agree at the same time there is no need of panic, its not going to help in any case.

>The alternative is that a typing mistake creates a logic mistake that can only be observed by test cases

Yes the test case needs to handle the return error

Panic causes problem to the whole operator which will side effect on the other controllers, IMO if something happens to a specific controller only affected CR's has the effect not the whole operator

internal/controller/driver_controller.go
nb-ohad: I agree, it is not allowed! but the solution is not to replace upper case letters with `-` but to replace them with the equivalent lower case letters:

So if we get the name: `Ohad_And_Madhu` as a name we need to change it to `ohad-and-madhu` and not to `-had--nd--madhu` which is what the proposed code does.

You can solve it in one of 2 ways:
1. change:  ```
name := r.generateName(suffix)
``` to ```name := strings.ToLower(r.generateName(suffix))``` and keep everything else the same
2. change ```regexp.Compile(`[^a-z0-9-]`)``` to ```re, err := regexp.Compile(`[^a-zA-Z0-9-]`)``` and also change ```result := re.ReplaceAllString(name, "-")``` to ```result := strings.ToLower(re.ReplaceAllString(name, "-"))```

Madhu-1: >The Driver "restinTin.cephfs.csi.ceph.com" is invalid: metadata.name: Invalid value: "restinTin.cephfs.csi.ceph.com": a lowercase RFC 1123 subdomain must consist of lower case alphanumeric characters, '-' or '.', and must start and end with an alphanumeric character (e.g. 'example.com', regex used for validation is '[a-z0-9]([-a-z0-9]*[a-z0-9])?(\.[a-z0-9]([-a-z0-9]*[a-z0-9])?)*')

Is it allowed to create a driver CR with an upper case letter? when that is allowed we should try to solve the problem.

nb-ohad: Ok, I think I understand what you are saying now. 
The CR name for a derived does not allow uppercase letters because it needs to comply with RFC 1123. In that case, I agree that the code above is ok. 

But can you please add a comment that states that the name we got from `generateName` is assumed to be RFC 1123 compatible?

Madhu-1: Done!

Issue #92, opened 2024-08-05
kubectl apply of make build-installer is failing


Madhu-1:
Error from server (Invalid): error when creating "dist/install.yaml": [CustomResourceDefinition.apiextensions.k8s.io](http://customresourcedefinition.apiextensions.k8s.io/) "[drivers.csi.ceph.io](http://drivers.csi.ceph.io/)" is invalid: spec.validation.openAPIV3Schema.properties[spec].properties[imageSet].x-kubernetes-validations[0].rule: Invalid value: apiextensions.ValidationRule{Rule:"self.?name.orValue(\"\") != \"\"", Message:"'.name' cannot be empty", MessageExpression:""}: compilation failed: ERROR: <input>:1:6: Syntax error: extraneous input '?' expecting IDENTIFIER
 | self.?name.orValue("") != ""
 | .....^
Error from server (Invalid): error when creating "dist/install.yaml": [CustomResourceDefinition.apiextensions.k8s.io](http://customresourcedefinition.apiextensions.k8s.io/) "[operatorconfigs.csi.ceph.io](http://operatorconfigs.csi.ceph.io/)" is invalid: spec.validation.openAPIV3Schema.properties[spec].properties[driverSpecDefaults].properties[imageSet].x-kubernetes-validations[0].rule: Invalid value: apiextensions.ValidationRule{Rule:"self.?name.orValue(\"\") != \"\"", Message:"'.name' cannot be empty", MessageExpression:""}: compilation failed: ERROR: <input>:1:6: Syntax error: extraneous input '?' expecting IDENTIFIER
 | self.?name.orValue("") != ""

PR #102, opened 2024-08-06
update validation rule for confimap name

Madhu-1:

updated the validation rule for configmap name to work for all kubernetes versions.

## Related issues ##

Fixes: #92

### DIFF
--- a/api/v1alpha1/driver_types.go
+++ b/api/v1alpha1/driver_types.go
@@ -273,7 +273,7 @@ type DriverSpec struct {
 	// A reference to a ConfigMap resource holding image overwrite for deployed
 	// containers
 	//+kubebuilder:validation:Optional
-	//+kubebuilder:validation:XValidation:rule=self.?name.orValue("") != "",message="'.name' cannot be empty"
+	//+kubebuilder:validation:XValidation:rule=self.name != "",message="'.name' cannot be empty"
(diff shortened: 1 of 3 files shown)


api/v1alpha1/driver_types.go
```diff
@@ -273,7 +273,7 @@ type DriverSpec struct {
 	// A reference to a ConfigMap resource holding image overwrite for deployed
 	// containers
 	//+kubebuilder:validation:Optional
-	//+kubebuilder:validation:XValidation:rule=self.?name.orValue("") != "",message="'.name' cannot be empty"
+	//+kubebuilder:validation:XValidation:rule=self.name != "",message="'.name' cannot be empty"
```
nb-ohad, line 276: @Madhu-1 Selef might not exist and be nil, that will throw the validation check.

nb-ohad (changes requested):

api/v1alpha1/driver_types.go
Madhu-1: can you specify what exact case doesn't work with sample yamls. when i tested it worked as expected

nb-ohad: Checked the new rule with `self = nil` on both 1.29 and 1.30 and it works.

nb-ohad (approved):

PR #104, opened 2024-08-06
Driver Controller: Ensure CSI Config Creation

nb-ohad:

If a driver CR exists but no client profile CR exists, any container that needs to mount the `ceph-csi-config` config map will fail. To solve the issue, the following behavior is added:

- Driver controller creates an empty `ceph-csi-config` config map if not already exist and adds an owner ref
- Driver controller watch for delete events for `ceph-csi-config` config map

### DIFF
--- a/internal/controller/clientprofile_controller.go
+++ b/internal/controller/clientprofile_controller.go
@@ -213,8 +213,8 @@ func (r *ClientProfileReconcile) reconcileCephConnection() error {
 
 	if needsUpdate, err := utils.ToggleOwnerReference(
 		!r.cleanUp,
-		&r.clientProfile,
 		&r.cephConn,
+		&r.clientProfile,
 		r.Scheme,
 	); err != nil {
 		r.log.Error(err, "Failed to toggle owner reference on CephConnection")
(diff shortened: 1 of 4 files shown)


nb-ohad:
/hold

This PR is rebased on top of #103 and should not be merged until #103 is merged

Madhu-1:
> * Driver controller creates an empty `ceph-csi-config` config map if not already exist and adds an owner ref
> * Driver controller watch for delete events for `ceph-csi-config` config map

Instead of watching for delete events can't we add the driver as the ownerRef if the configmap already exists and we don't need to consider delete events?

nb-ohad:
/unhold

#103 merged, @Madhu-1 can you please add this to your review queue?

nb-ohad:
> > * Driver controller creates an empty `ceph-csi-config` config map if not already exist and adds an owner ref
> > * Driver controller watch for delete events for `ceph-csi-config` config map
> 
> Instead of watching for delete events can't we add the driver as the ownerRef if the configmap already exists and we don't need to consider delete events?

We are already adding the driver as an owner ref. But there is still a need to watch for delete events for the case where the configmap goes away and need to be recreated. The delete filter makes sure that  we are not queuing an based on create or update events

Madhu-1:
> > > * Driver controller creates an empty `ceph-csi-config` config map if not already exist and adds an owner ref
> > > * Driver controller watch for delete events for `ceph-csi-config` config map
> > 
> > 
> > Instead of watching for delete events can't we add the driver as the ownerRef if the configmap already exists and we don't need to consider delete events?
> 
> We are already adding the driver as an owner ref. But there is still a need to watch for delete events for the case where the configmap goes away and need to be recreated. The delete filter makes sure that we are not queuing an based on create or update events

Sorry am failing to understand it as we are adding the owner ref if the configmap already exists how it will go away if the driver is not deleted? Do you mean the use will manually remove the configmap by hand?

nb-ohad:
Yes.
It can be faulty code in external automation, or some bug in rook or client, or a mistake by an admin, or someone just overwrote the entire ownerRef section instead of adding to it. 

If a controller takes the responsibility to move a system toward a certain actual state it also needs to take the responsibility to readjust if that actual state has changed (for any reason)

Madhu-1 (approved):

PR #105, opened 2024-08-07
remove controller runtime dependency

Madhu-1:

we have seen the issues in other projects where API has controller runtime as a dependency though the main project importing the operator doesn't have the dependency on controller runtime directly now they need to wait for it to update to other kubernetes versions as well.

https://github.com/ceph/ceph-csi-operator/pull/94#issuecomment-2272512200


Madhu-1:
@leelavg ^^

Madhu-1:
If all agree on this one, am going to test it later

leelavg:
> @leelavg ^^
- able to build client-op w/ these changes. thanks.

nb-ohad (approved):

nb-ohad:
I would suggest having a different approach here. We can create our own interface on top of Register and use it inside the API Package.

The can be considered for the future

Issue #106, opened 2024-08-07
Operator needs service account csi-rbd-nodeplugin-sa but rolebindings includes as ceph-csi-operator-csi-rbd-nodeplugin-sa


parth-gr:
# Describe the bug #

currently, the operator code needs service account `csi-rbd-nodeplugin-sa` sa but rolebindings include as `ceph-csi-operator-csi-rbd-nodeplugin-sa`,

So either we need, 

1) Change the service account to add prefix `ceph-csi-operator`

2) Remove prefix `ceph-csi-operator` from the roles and rolebindings and clusterrolebindings.

Need a discussion on what change we should pick, and collect different views.

# Environment details #

- Image/version of ceph-csi-operator :
- Image/version of Ceph-CSI driver :
- Kubernetes cluster version :
- Ceph cluster version :

# Steps to reproduce #

1. Setup details: '...'
1. Deployment to trigger the issue '....'
1. See error

# Actual results #

Describe what happened

# Expected behavior #

# Logs #

# Additional context #

For example:

Any existing bug report which describe about the similar issue/behavior

travisn:
The service account names as defined currently, for example [csi-cephfs-ctrlplugin-sa](https://github.com/ceph/ceph-csi-operator/blob/main/config/rbac/csi_cephfs_ctrlplugin_service_account.yaml#L4), are nice and simple. Can we keep the simple names as the default? Why would we need a `ceph-csi-operator` prefix on the name? This would make a name of `ceph-csi-operator-csi-cephfs-ctrlplugin-sa` which seems unnecessarily long and redundant as the default. 

If someone wants to include a second driver in the namespace, they could set their own service account name and configure the [serviceAccountName](https://github.com/ceph/ceph-csi-operator/blob/main/api/v1alpha1/driver_types.go#L97-L99) to use an alternate. Isn't that the purpose of that field already to allow setting a non-default option? So we can keep the default option without the prefix. Let's optimize for the common case of one driver in the namespace.

Madhu-1:
@travisn This is because of the way how we are installing, if we install using the kustomize this is expected.

Madhu-1:
@parth-gr note we do have the option to specify the service account names for both the controller plugin and node-plugin in the Driver CR and the OperatorConfig CR which you can use.

Madhu-1:
Reopening as now its not fixed but a way was provided to generate the CSI RBAC with expected names but we still need to fix the prefix part for the operator

[...]

Issue #87, opened 2024-08-05
Driver's service account is created in system namespace


subhamkrai:
driver's service account should not be created in the system namespace otherwise replica set fail to scale up with error
```
  Warning  FailedCreate  19s (x16 over 3m3s)  replicaset-controller  Error creating: pods "rook-ceph.cephfs.csi.ceph.com-ctrlplugin-6ffb4b774-" is forbidden: error looking up service account rook-ceph/csi-cephfs-ctrlplugin-sa: serviceaccount "csi-cephfs-ctrlplugin-sa" not found

````

Madhu-1:
@subhamkrai This is a manual step for the consumer, its consumer's responsibility to create the cephcsi required RBAC's in the same namespace where they are creating the driver CR's. nothing can be done in this operator to fix it. maybe we can add a make file target/document the step for this nothing more than that.

subhamkrai:
document or make command sounds good

PR #107, opened 2024-08-08
fix csi rbac in kustomize

Madhu-1:
Moving the csi RBAC to a new folder can be built and applied differently and we dont need to have the same namePrefix and namespacePrefix for these files as this is not required to be deployed in the same namespace as the operator.

fixes: #87
fixes: #106


Madhu-1:
Note:- This still needs some changes to work for both downstream and upstream. putting this one hold for now

Makefile
```diff
@@ -4,6 +4,14 @@ REGISTRY_NAMESPACE ?= cephcsi
 IMAGE_TAG ?= latest
 IMAGE_NAME ?= ceph-csi-operator
 
+# Use different name prefix and namespace prefix for kustomize
+KUSTOMIZE_NAME_PREFIX ?= ceph-csi-operator-
+KUSTOMIZE_NAMESPACE_PREFIX ?= system
+KUSTOMIZE_CSI_NAME_PREFIX ?= csi-
+KUSTOMIZE_CSI_NAMESPACE_PREFIX ?= system
```
nb-ohad, line 11: This is a suffix not a prefix

Makefile
```diff
@@ -121,17 +129,39 @@ docker-buildx: ## Build and push docker image for the manager for cross-platform
...
   ignore-not-found = false
 endif
 
+PHONY: update-prefix
+update-prefix: backup-prefix ## Update the kustomization.yaml with the new prefix
+	@echo "Updating prefixes in kustomization.yaml files..."
+	sed -i.bak "s/namePrefix: .*/namePrefix: ${KUSTOMIZE_NAME_PREFIX}/" config/default/kustomization.yaml
+	sed -i.bak "s/namespace: .*/namespace: ${NAMESPACE_PREFIX}/" config/default/kustomization.yaml
+	sed -i.bak "s/namePrefix: .*/namePrefix: ${KUSTOMIZE_CSI_NAME_PREFIX}/" config/csi-rbac/kustomization.yaml
+	sed -i.bak "s/namespace: .*/namespace: ${CSI_NAMESPACE_PREFIX}/" config/csi-rbac/kustomization.yaml
```
nb-ohad, line 151: Why use a sed? The entire point of using kustomize is to have no need for text replacement operations.
If you need to apply additional kaustomize options based on you own values, you need to write a new file with the required changes then apply it as the root of your kustomize operation

Makefile
```diff
@@ -121,17 +129,39 @@ docker-buildx: ## Build and push docker image for the manager for cross-platform
...
+
+.PHONY: revert-prefix
+revert-prefix: ## Revert the kustomization.yaml to original state
+	@echo "Reverting prefixes in kustomization.yaml files..."
+	@if [ -f config/default/kustomization.yaml.bak ]; then mv config/default/kustomization.yaml.bak config/default/kustomization.yaml; fi
+	@if [ -f config/csi-rbac/kustomization.yaml.bak ]; then mv config/csi-rbac/kustomization.yaml.bak config/csi-rbac/kustomization.yaml; fi
```
nb-ohad, line 157: This is unnecessary if we follow the kustomize way of applying patches.

Makefile
```diff
@@ -121,17 +129,39 @@ docker-buildx: ## Build and push docker image for the manager for cross-platform
...
+
+.PHONY: backup-prefix
+backup-prefix: ## Backup the kustomization.yaml files
+	@echo "Backing up kustomization.yaml files..."
+	@cp config/default/kustomization.yaml config/default/kustomization.yaml.bak
+	@cp config/csi-rbac/kustomization.yaml config/csi-rbac/kustomization.yaml.bak
+
```
nb-ohad, line 164: This is unnecessary if we follow the kustomize way of applying patches.

config/csi-rbac/csi_cephfs_ctrlplugin_cluster_role.yaml
nb-ohad, line 1: Why are we getting rid of the csi prefix? I am not against it, I am just asking why?

Also if we are, we should also rename the file, as I think that having the file name similar to the object name is beneficial 

Also, we move the file to a different folder?

config/csi-rbac/kustomization.yaml
```diff
@@ -0,0 +1,30 @@
+# Adds namespace to all resources.
+namespace: csi-system
```
nb-ohad, line 2: default ns must match the operator default ns, if not it will be unusable when generating without customization

nb-ohad (changes requested):

config/csi-rbac/kustomization.yaml
Madhu-1: This is following the standard what kustomization says, the namespace should have the name suffix

config/csi-rbac/csi_cephfs_ctrlplugin_cluster_role.yaml
Madhu-1: > Why are we getting rid of the csi prefix? I am not against it, I am just asking why?

This is to adjust the name prefix for what we have today, am okay to revert it as well, This keeps the things simple

> Also if we are, we should also rename the file, as I think that having the file name similar to the object name is beneficial

Nothing is changed in name, afaik everything remains the same only moved to a new folder.

> Also, we move the file to a different folder?

This is to provide flexibility to generate the RBAC only for CSI as it need to be generated per driver or per namespace as per user customization.

Makefile
Madhu-1: This is the flexibility to provide kustomization in the makefile nothing else, if users want to run customization manually they can do it.

Makefile
Madhu-1: same reason as below.

Makefile
Madhu-1: This is to have flexibility where user need not to worry about customization and just make use of makefile, am fine to drop it as well if not required.

Madhu-1: Thanks @nb-ohad for suggesting `kustomize edit set nameprefix` i will remove sed

Madhu-1 (commented):
Am not fixing the variable to set the prefix in deployment and code changes in this PR.

Makefile
```diff
@@ -121,11 +125,16 @@ docker-buildx: ## Build and push docker image for the manager for cross-platform
 	rm Dockerfile.cross
 
 .PHONY: build-installer
-build-installer: manifests generate kustomize ## Generate a consolidated YAML with CRDs and deployment.
+build-installer: manifests generate kustomize build-csi-rbac ## Generate a consolidated YAML with CRDs and deployment.
```
nb-ohad, line 128: You don't need this to do this.
 You can, instead, add the new csi folder into `default/kusomizatoin.yaml` as all other folders.

Makefile
```diff
@@ -144,10 +153,13 @@ uninstall: manifests kustomize ## Uninstall CRDs from the K8s cluster specified
 deploy: manifests kustomize ## Deploy controller to the K8s cluster specified in ~/.kube/config.
 	cd config/manager && $(KUSTOMIZE) edit set image controller=${IMG}
 	$(KUSTOMIZE) build config/default | $(KUBECTL) apply -f -
+	$(KUSTOMIZE) build config/csi-rbac | $(KUBECTL) apply -f -
```
nb-ohad, line 156: You don't need this to do this.
You can, instead, add the new csi folder into default/kusomizatoin.yaml as all other folders.

Makefile
```diff
@@ -144,10 +153,13 @@ uninstall: manifests kustomize ## Uninstall CRDs from the K8s cluster specified
...
+
 
 .PHONY: undeploy
 undeploy: kustomize ## Undeploy controller from the K8s cluster specified in ~/.kube/config. Call with ignore-not-found=true to ignore resource not found errors during deletion.
 	$(KUSTOMIZE) build config/default | $(KUBECTL) delete --ignore-not-found=$(ignore-not-found) -f -
+	$(KUSTOMIZE) build config/csi-rbac | $(KUBECTL) delete --ignore-not-found=$(ignore-not-found) -f -
```
nb-ohad, line 162: You don't need this to do this.
You can, instead, add the new csi folder into default/kusomizatoin.yaml as all other folders.

config/csi-rbac/csi_cephfs_ctrlplugin_cluster_role.yaml
nb-ohad, line 1: @Madhu-1 can you please remove the `csi` prefix from the file name to align with the internal resource name changes

nb-ohad (changes requested):

nb-ohad (approved):

PR #116, opened 2024-08-12
Configuration: Kustomisation enhancements

nb-ohad:

- Make commands that use kustomize no longer edit the local config folder and have zero side effects on local repo files
- Driver controller now accepts a service account prefix from an env var injected into the deployment via kustomize
- Enable setting of env vars before `build-installer`, `build-csi-rbac`, and `deploy` make target to allow kustomization. Example:

```NAME_PREFIX=\<prefix\> NAMESPACE=\<target_namspace\> make deploy``

This change also ensures that both `build-installer` and `deploy` create
RBAC, CSI RBAC, and SA accounts that allow the operator to run without
any need for manual changes!

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -4,9 +4,9 @@ REGISTRY_NAMESPACE ?= cephcsi
 IMAGE_TAG ?= latest
 IMAGE_NAME ?= ceph-csi-operator
 
-# Use different name prefix and namespace prefix for csi rbac kustomize
-CSI_RBAC_NAME_PREFIX ?= ceph-csi-operator-
-CSI_RBAC_NAMESPACE ?= $(CSI_RBAC_NAME_PREFIX)system
+# Allow customization of the name prefix and/or namespace
+NAME_PREFIX ?= ceph-csi-operator-
+NAMESPACE ?= $(NAME_PREFIX)system
 
 IMG ?= $(IMAGE_REGISTRY)/$(REGISTRY_NAMESPACE)/$(IMAGE_NAME):$(IMAGE_TAG)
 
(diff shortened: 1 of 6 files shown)

### CODE REVIEW

File: Makefile [hunk 1]
```diff
@@ -31,6 +31,44 @@ CONTAINER_TOOL ?= docker
 SHELL = /usr/bin/env bash -o pipefail
 .SHELLFLAGS = -ec
 
+# Define the content of the temporary top-most kustomize overlay for the
```
iamniting, line 34: @nb-ohad I think we can make is too simple just with one command

```
cd config/manager && kustomize edit add patch --name controller-manager --kind Deployment --patch '[{"op": "replace", "path": "/spec/template/spec/containers/0/env/1/value", "value": "prefix"}]'
```

Also please add the `CSI_SERVICE_ACCOUNT_PREFIX` env variable and its empty default value in the manager.

Pls look at https://github.com/iamniting/ceph-csi-operator/commit/63c09a52b154c9acc4bb518134bab931b5707306

config/manager/kustomization.yaml
```diff
@@ -2,7 +2,3 @@ resources:
 - manager.yaml
 apiVersion: kustomize.config.k8s.io/v1beta1
 kind: Kustomization
-images:
-- name: controller
-  newName: quay.io/cephcsi/ceph-csi-operator
-  newTag: latest
```
iamniting, line 8: Lets not remove it from here.

iamniting (changes requested):

Makefile
nb-ohad: @iamniting 
Thank you for the suggestion, but I think you missed the point. We should no edit existing Kustomize files while trying to build. These files are tracked by the repo, which means that any change to them is persistent. 
The more safe way to influence the customization is to build an overlay file and introduce your changes there.

config/manager/kustomization.yaml
nb-ohad: We cannot, changes are cascading which means any patch here "overwrites" any patch coming from a higher overlay. I was chasing this bug almost 1.5 hours before I saw you committed this change some time ago.
As a rule of thumb, if we need customization on top of any persistent repo file it should come from an overlay.

Makefile
iamniting: I really don't agree with you here that we should not change the existing Kustomize files. Kustomize files should be changed whenever required. We are just introducing an extra layer on top, which I believe is not really necessary.

config/manager/kustomization.yaml
iamniting: We dont need an higher overlay at all IMO.

Makefile
nb-ohad: @iamniting Kustomize files should not be personalized locally and then pushed. 
You already did that, I assume by mistake, by running the make target and then committing the changes. That is what created the container image patch inside the manager folder.  

The point is that we should like to have the kustomizes files static and not influenced by local commands running on local branches/forks

config/manager/kustomization.yaml
nb-ohad: I would say the opposite! we should not touch any of the existing bases/overlays with local changes

Madhu-1 (approved):

Makefile
nb-ohad: As this is not about the correctness of the solution but about the approach (which is somewhat a matter of opinion),  I am resolving this so we could merge, to unblock us. We can continue the discussion after merging and if needed we can always pivot with a new PR on top

config/manager/kustomization.yaml
nb-ohad: As this is not about the correctness of the solution but about the approach (which is somewhat a matter of opinion),  I am resolving this so we could merge, to unblock us. We can continue the discussion after merging and if needed we can always pivot with a new PR on top

PR #125, opened 2024-08-20
Generate multifile installer

Madhu-1:

This PR add a new makefile target to generate the separate yamls for the operator installation and also updates the github action to verify that no changes are manually done to the deploy folder as well by running the `make build-installer` and `make build-multifile-installer`

Do you have any questions?

No

Yes

No

For example:

## Related issues ##

Fixes: #issue_number

## Future concerns ##

**Checklist:**

  guidelines in the [developer
* [ ] [Pending release


Makefile
```diff
@@ -168,17 +185,26 @@ docker-buildx: ## Build and push docker image for the manager for cross-platform
...
 .PHONY: build-installer
 build-installer: manifests generate kustomize ## Generate a consolidated YAML with CRDs and deployment.
-	mkdir -p build dist
+	mkdir -p build deploy/all-in-one
 	cd build && echo "$$BUILD_INSTALLER_OVERLAY" > kustomization.yaml
-	$(KUSTOMIZE) build build > dist/install.yaml
+	$(KUSTOMIZE) build build > deploy/all-in-one/install.yaml
+	rm -rf build
+
+.PHONY: build-custom-installer
```
nb-ohad, line 193: Just adding another example for the case to change then name:
```suggestion
.PHONY: build-multifile-installer
```

config/manager/kustomization.yaml
```diff
@@ -2,3 +2,8 @@ resources:
 - manager.yaml
 apiVersion: kustomize.config.k8s.io/v1beta1
 kind: Kustomization
+patches:
+# Protect the /metrics endpoint by putting it behind auth.
+# If you want your controller-manager to expose the /metrics
+# endpoint w/o any authn/z, please comment the following line.
+- path: manager_auth_proxy_patch.yaml
```
nb-ohad, line 9: I am not sure about this change. There is a reason why the generate code decided to put it in the top most overlay and not in the manager's overlay

deploy/kubernetes/crd.yaml
nb-ohad, line 1: Let's rename the folder to `deploy/multifile/*.yaml` 
Asking as the all-in-one is also a kubernetes deployer

nb-ohad (changes requested):

Makefile
```diff
@@ -55,12 +55,29 @@ images:
...
 
+# Define the content of the temporary top-most kustomize overlay for the
+# build-custom-installer.
+define BUILD_CUSTOM_OVERLAY
+$(COMMON_OVERLAY)
+resources:
+- ../config/rbac
+- ../config/manager
+endef
+export BUILD_CUSTOM_OVERLAY
```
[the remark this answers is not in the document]
Madhu-1: the idea is to have a single CRD file and operator deployment and its RBAC in one file and csi back in another file. that the reason we see two resources expect csi and CRD, let me know if you have a suggestion to do it without it.

config/manager/kustomization.yaml
Madhu-1: i have  seen repo's moving this as per their requirements/customizations, and I don't see any problem in moving this to the manager as well, If you see it as a problem let me know what problem it can cause and alternative suggestion as well

nb-ohad: @Madhu-1 I would say, let's not move it from it original location if there is no real reason to do so

Makefile
```diff
@@ -55,12 +55,29 @@ images:
...
 
+# Define the content of the temporary top-most kustomize overlay for the
+# build-multifile-installer.
+define BUILD_MULTIFILE_OVERLAY
+$(COMMON_OVERLAY)
+resources:
+- ../config/rbac
+- ../config/manager
+endef
+export BUILD_MULTIFILE_OVERLAY
```
nb-ohad, line 79: Instead of using a new var that uses the common part via env avr injection. Can we just create the file from the common overlay and use the kustimize way to add the resources section in each build target? (using `kustimize edit` commands)

nb-ohad (changes requested):

Makefile
Madhu-1: updated to use edit as per discussion

config/manager/kustomization.yaml
Madhu-1: added a comment on top of the files as per discussion

Issue #133, opened 2024-09-09
make test fails on fresh clone


obnoxxx:
# Describe the bug #

On my Fedora 40 system, `make test`fails with a panic on a fresh `main` clone:

```console

$ make test
/home/obnox/go/src/github.com/ceph/ceph-csi-operator/bin/controller-gen-v0.14.0 rbac:roleName=manager-role crd webhook paths="./..." output:crd:artifacts:config=config/crd/bases
/home/obnox/go/src/github.com/ceph/ceph-csi-operator/bin/controller-gen-v0.14.0 object:headerFile="hack/boilerplate.go.txt" paths="./..."
go fmt ./...
go vet ./...
Downloading sigs.k8s.io/controller-runtime/tools/setup-envtest@release-0.17
go: downloading sigs.k8s.io/controller-runtime/tools/setup-envtest v0.0.0-20240812162837-9557f1031fe4
go: downloading sigs.k8s.io/controller-runtime v0.17.6
go: downloading github.com/spf13/afero v1.6.0
go: downloading go.uber.org/zap v1.26.0
go: downloading sigs.k8s.io/yaml v1.3.0
go: downloading go.uber.org/multierr v1.10.0
KUBEBUILDER_ASSETS="/home/obnox/go/src/github.com/ceph/ceph-csi-operator/bin/k8s/1.29.0-linux-amd64" go test $(go list ./... | grep -v /e2e) -coverprofile cover.out
	github.com/ceph/ceph-csi-operator/cmd		coverage: 0.0% of statements
	github.com/ceph/ceph-csi-operator/test/utils		coverage: 0.0% of statements
	github.com/ceph/ceph-csi-operator/internal/utils		coverage: 0.0% of statements
panic: Required OPERATOR_NAMESPACE environment variable is either missing or empty

goroutine 1 [running]:
github.com/ceph/ceph-csi-operator/internal/controller.init.func1(...)
	/home/obnox/go/src/github.com/ceph/ceph-csi-operator/internal/controller/defaults.go:63
github.com/ceph/ceph-csi-operator/internal/utils.Call[...](...)
	/home/obnox/go/src/github.com/ceph/ceph-csi-operator/internal/utils/core.go:121
github.com/ceph/ceph-csi-operator/internal/controller.init()
	/home/obnox/go/src/github.com/ceph/ceph-csi-operator/internal/controller/defaults.go:60 +0x485
[...]

PR #134, opened 2024-09-11
Makefile: fix panic in make test.

obnoxxx:

This stops `make test` from failing with a panic of  the operator by making sure OPERATOR_NAMESPACE is set in the shell environment.

It does not fix make test completely though  but lets it at least pass the first obstacle.

Do you have any questions? - no.

Is the change backward compatible? - should be.

Are there concerns around backward compatibility? - no.

## Related issues ##

Resolves: #133 

## Future concerns ##

* `make test` is still failing. this should be fixed
* `make test`is not currently run in the CI. It should be.
* 

**Checklist:**

  guidelines in the [developer
* [ ] [Pending release

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -134,7 +134,7 @@ build: manifests generate fmt vet ## Build manager binary.
 
 .PHONY: run
 run: manifests generate fmt vet ## Run a controller from your host.
-	go run ./cmd/main.go
+	OPERATOR_NAMESPACE="$(shell echo $${OPERATOR_NAMESPACE:=$(NAMESPACE)})" go run ./cmd/main.go
 
 # If you wish to build the manager image targeting other platforms you can use the --platform flag.
 # (i.e. docker build --platform linux/arm64). However, you must enable docker buildKit for it.
--- a/Makefile
+++ b/Makefile
@@ -111,7 +111,7 @@ vet: ## Run go vet against code.
 
 .PHONY: test
 test: manifests generate fmt vet envtest ## Run tests.
-	KUBEBUILDER_ASSETS="$(shell $(ENVTEST) use $(ENVTEST_K8S_VERSION) --bin-dir $(LOCALBIN) -p path)" go test $$(go list ./... | grep -v /e2e) -coverprofile cover.out
+	OPERATOR_NAMESPACE="$(shell echo $${OPERATOR_NAMESPACE:=$(NAMESPACE)})" KUBEBUILDER_ASSETS="$(shell $(ENVTEST) use $(ENVTEST_K8S_VERSION) --bin-dir $(LOCALBIN) -p path)" go test $$(go list ./... | grep -v /e2e) -coverprofile cover.out
 
 # Utilize Kind or modify the e2e tests to load the image locally, enabling compatibility with other vendors.
 .PHONY: test-e2e  # Run the e2e tests against a Kind k8s instance that is spun up.


Makefile
```diff
@@ -7,6 +7,12 @@ IMAGE_NAME ?= ceph-csi-operator
 # Allow customization of the name prefix and/or namespace
 NAME_PREFIX ?= ceph-csi-operator-
 NAMESPACE ?= $(NAME_PREFIX)system
+OPERATOR_NAMESPACE = $(shell echo "${$OPERATOR_NAMESPACE}")
```
Madhu-1, line 10: IMO i dont see a need for another variable for operator namespace we can just use `NAMESPACE` in make test

obnoxxx: @Madhu-1  wrote:
> IMO i dont see a need for another variable for operator namespace we can just use `NAMESPACE` in make test

Agreed, but when I tried to set it directly in the test target definition, I struggled with makefile vs shell syntax because I want to set it conditionally: only when the var is not set in the outside shell env.

obnoxxx: fixed to set it unconditionally

Madhu-1: if someone wants to set the NAMESPACE for the operator to run the test they can set the `NAMESPACE` that we already have, we should not two different variables to serve the same purpose to set the NAMESPACE for the operator.

Makefile
```diff
@@ -8,6 +8,8 @@ IMAGE_NAME ?= ceph-csi-operator
 NAME_PREFIX ?= ceph-csi-operator-
 NAMESPACE ?= $(NAME_PREFIX)system
 
+
+
```
Madhu-1, line 12: perfect can you please remove this extra two new lines?

obnoxxx: gone

Madhu-1 (approved):

Makefile
```diff
@@ -111,7 +111,7 @@ vet: ## Run go vet against code.
 
 .PHONY: test
 test: manifests generate fmt vet envtest ## Run tests.
-	KUBEBUILDER_ASSETS="$(shell $(ENVTEST) use $(ENVTEST_K8S_VERSION) --bin-dir $(LOCALBIN) -p path)" go test $$(go list ./... | grep -v /e2e) -coverprofile cover.out
+	OPERATOR_NAMESPACE="$(NAMESPACE)" KUBEBUILDER_ASSETS="$(shell $(ENVTEST) use $(ENVTEST_K8S_VERSION) --bin-dir $(LOCALBIN) -p path)" go test $$(go list ./... | grep -v /e2e) -coverprofile cover.out
```
nb-ohad, line 114: This will always overwrite the OPERATOR_NAMESPACE env var, which means that we are breaking existing functionality.

Consider the following command:
```
export OPERATOR_NAMESPACE=my-namespace
make test
```

or even
```
OPERATOR_NAMESPACE=my-namespace make test
```

In both cases, the operator will not get tested in "my-namespace" which is a breaking of the existing behavior

nb-ohad (changes requested):

Makefile
Madhu-1: @nb-ohad i dont think anything is working as existing functionality, what is the use of allowing a different name other than the `NAMESPACE` one?

obnoxxx: @nb-ohad  wrote:
> This will always overwrite the OPERATOR_NAMESPACE env var, which means that we are breaking existing functionality.
> 
> Consider the following command:
> 
> ```
> export OPERATOR_NAMESPACE=my-namespace
> make test
> ```
> 
> or even
> 
> ```
> OPERATOR_NAMESPACE=my-namespace make test
> ```
> 
> In both cases, the operator will not get tested in "my-namespace" which is a breaking of the existing behavior

Excellent point!

This is exactly why I wanted to set the env var only conditionally, when not already set. I used a new Makefile variable to circumvent syntax issues.

Upon @Madhu-1 's request, I changed it to set unconditionally.

I can try again to set it conditionally in the  test target definition without introducing the extra variable.

nb-ohad: @Madhu-1 I am using is locally when running multiple make test at the same time

Madhu-1: why not set `NAMSPACE` if you are using makefile to run the tests? i don't like to have specific variables for the test as we already have a variable for the operator namespace(kustomize namespace).

obnoxxx: @Madhu-1  wrote:
> why not set `NAMSPACE` if you are using makefile to run the tests? i don't like to have specific variables for the test as we already have a variable for the operator namespace(kustomize namespace).


do you mean using `make NAMESPACE=foo test` ?

I just confirmed that this works with the current patch.

Also `make NAMESPACE="" test`correctly triggers the panic. :) 

@nb-ohad  is this sufficient for your purpose?

nb-ohad: @Madhu-1 NAMESPACE affects other parts of the dependent make targets, I just want to let the operator know which namespace it is running on.

I am ok to defaulting to NAMESPACE if there is no OPERATOR_NAMESPACE defined. But I think the other way around is too limiting

obnoxxx: The latest update only sets the OPERATOR_NAMESPACE env var  (only in the test target) if it is not already set outside of make.
It avoids a new Makefile var.

Overriding it with

`OPERATOR_NAMESPACE=foo make test` and `make NAMESPACE=foo test` both weork. 

@Madhu-1 , @nb-ohad  - wdyt?

nb-ohad: LGTM

nb-ohad:
@obnoxxx Please consider that we will need a similar patch for `make run`

Madhu-1:
> @obnoxxx Please consider that we will need a similar patch for `make run`

@obnoxxx can you please the same for make run as well as part of this PR (different commit)

Madhu-1 (approved):

Madhu-1:
@obnoxxx you have requested for changes, PTAL

Madhu-1:
> @obnoxxx you have requested for changes, PTAL

sorry for my mistake, @nb-ohad PTAL

nb-ohad (approved):

Makefile
```diff
@@ -111,7 +111,7 @@ vet: ## Run go vet against code.
 
 .PHONY: test
 test: manifests generate fmt vet envtest ## Run tests.
-	KUBEBUILDER_ASSETS="$(shell $(ENVTEST) use $(ENVTEST_K8S_VERSION) --bin-dir $(LOCALBIN) -p path)" go test $$(go list ./... | grep -v /e2e) -coverprofile cover.out
+	OPERATOR_NAMESPACE="$(shell echo $${OPERATOR_NAMESPACE:=$(NAMESPACE)})" KUBEBUILDER_ASSETS="$(shell $(ENVTEST) use $(ENVTEST_K8S_VERSION) --bin-dir $(LOCALBIN) -p path)" go test $$(go list ./... | grep -v /e2e) -coverprofile cover.out
```
BlaineEXE, line 114: FYI, I'm pretty sure that this is more verbose than necessary. I am pretty sure that this shorter and easier to read line is equivalent

```
OPERATOR_NAMESPACE:="$${OPERATOR_NAMESPACE:-$(NAMESPACE)" KUBEBUILDER_ASSETS=...
````

obnoxxx: Thanks for your review and suggestion, @BlaineEXE  !

I am pretty sure I started trying exactly what you suggest,  and it did not work for me, but I will try again.

obnoxxx: @BlaineEXE  I tried again and it works - thanks!

See #138

PR #148, opened 2024-09-26
doc: add quick-start guide

Madhu-1:

added a quick start guide for the minimal installation

**Checklist:**

  guidelines in the [developer
* [ ] [Pending release

### DIFF
--- a/docs/quick-start.md
+++ b/docs/quick-start.md
@@ -1,0 +1,158 @@
+- [Quick Start Guide for Ceph-CSI-Operator](#quick-start-guide-for-ceph-csi-operator)
+  - [1. Prerequisites](#1-prerequisites)
+  - [2. Install the Ceph-CSI Operator](#2-install-the-ceph-csi-operator)
+    - [3. Deploy Ceph-CSI Drivers](#3-deploy-ceph-csi-drivers)
+      - [3.1 Deploy the RBD Driver](#31-deploy-the-rbd-driver)
+      - [3.2 Deploy the CephFS Driver](#32-deploy-the-cephfs-driver)
+    - [3.3 Deploy the Ceph-NFS Driver](#33-deploy-the-ceph-nfs-driver)
+  - [4. Verify Installation](#4-verify-installation)
+  - [5. Create CephConnection](#5-create-cephconnection)
+  - [6. Create ClientProfile](#6-create-clientprofile)
+
+# Quick Start Guide for Ceph-CSI-Operator
+
+## 1. Prerequisites
+
+Before deploying the Ceph-CSI-Operator, ensure the following requirements are met:
+
+- A Kubernetes cluster ([supported version](https://kubernetes.io/releases/) recommended)
+- Ceph cluster ([supported version](https://docs.ceph.com/en/latest/releases/) recommended)
+- `kubectl` CLI installed
+
+**Note:** In this guide, we will use minimal configurations to deploy the Ceph-CSI-Operator and drivers. You can customize the configurations as per your requirements.
+
+## 2. Install the Ceph-CSI Operator
+
+```console
+kubectl create -f deploy/all-in-one/install.yaml
+```
+
+verify the installation:
+
+```bash
+kubectl get pods -n ceph-csi-operator-system
+NAME                                                    READY   STATUS    RESTARTS   AGE
+ceph-csi-operator-controller-manager-67d45fd9ff-zgst7   2/2     Running   0          40s
+```
+
+### 3. Deploy Ceph-CSI Drivers
+
+Once the operator is installed, deploy the Ceph-CSI drivers:
+
+#### 3.1 Deploy the RBD Driver
+
+```console
+echo '
+apiVersion: csi.ceph.io/v1alpha1
+kind: Driver
+metadata:
+  name: rbd.csi.ceph.com
+  namespace: ceph-csi-operator-system
+' | kubectl create -f -
+```
+
+#### 3.2 Deploy the CephFS Driver
+
+```console
+echo '
+apiVersion: csi.ceph.io/v1alpha1
+kind: Driver
+metadata:
+  name: cephfs.csi.ceph.com
+  namespace: ceph-csi-operator-system
+' | kubectl create -f -
+```
+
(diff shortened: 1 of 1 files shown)


Madhu-1:
@nb-ohad PTAL

docs/quick-start.md
```diff
@@ -0,0 +1,130 @@
...
+Create an OperatorConfig CR to configure the Ceph-CSI Operator:
+
+```console
+echo '
+apiVersion: csi.ceph.io/v1alpha1
+kind: OperatorConfig
+metadata:
+  name: ceph-csi-operator-config
+  namespace: ceph-csi-operator-system
+' | kubectl create -f -
```
nb-ohad, line 49: If we are going for minimal installation, why do we need an operator config instance?

docs/quick-start.md
```diff
@@ -0,0 +1,130 @@
...
+rbd.csi.ceph.com-ctrlplugin-6965dcfdb8-w88kn            5/5     Running   0          4m35s
+rbd.csi.ceph.com-nodeplugin-lnm4n                       2/2     Running   0          4m35s
+```
+
+## 6. Clean Up Resources
+
+To clean up the resources, delete the OperatorConfig CR, drivers:
+
+```console
+kubectl delete operatorconfig ceph-csi-operator-config -n ceph-csi-operator-system
```
nb-ohad, line 113: Continued from the last comment, for a minimal installation example we don't need an operator config CR

docs/quick-start.md
nb-ohad, line 1: If we want a minimal installation that can actually serve storage, we need the user to create at least one ceph connection and at least one client profile. Both are missing from this quick start guide.

nb-ohad (changes requested):

docs/quick-start.md
Madhu-1: it was added to deploy the CR with minimal values and no values and the user can update it later as per requirement. Removed it now

Madhu-1:
@nb-ohad PTAL

docs/quick-start.md
```diff
@@ -0,0 +1,160 @@
...
+echo '
+apiVersion: csi.ceph.io/v1alpha1
+kind: ClientProfile
+metadata:
+  name: storage
+  namespace: ceph-csi-operator-system
+spec:
+  cephConnectionRef:
+    name: ceph-connection
+  cephFS:
```
iPraveenParihar, line 126:
```suggestion
  cephFs:
```

```
$ k explain clientprofiles.csi.ceph.io.spec.cephFs
GROUP:      csi.ceph.io
KIND:       ClientProfile
VERSION:    v1alpha1

FIELD: cephFs <Object>
```

obnoxxx (commented):
I am surprised that the guide does not mention building the operator. Even if this is making use of published images, I think developers want to test locally  built images with their local changes.

Madhu-1:
> I am surprised that the guide does not mention building the operator. Even if this is making use of published images, I think developers want to test locally built images with their local changes.

That will be a follow-up to add/update the document on how to test the local changes.

obnoxxx (commented):
@Madhu-1  thanks for adding quickstart guide. Very useful!

A general suggestion:
Instead of the echo and using `kubectl create -f -`, wouldn't it be convenient to have example yaml files for these items (rbd driver, cephfs driver, client connection, client profile, etc)? this would probably be as follow-up PR.

obnoxxx (changes requested):
I tried the steps in the guide and they all work with one exception:

Creation of the `clientProfile` fails like this:

```console
Error from server (BadRequest): error when creating "STDIN": ClientProfile in version "v1alpha1" cannot be handled as a ClientProfile: strict decoding error: unknown field "spec.cephFs.subvolumeGroup"
```

docs/quick-start.md
```diff
@@ -0,0 +1,160 @@
...
+  - [6. Create ClientProfile](#6-create-clientprofile)
+
+# Quick Start Guide for Ceph-CSI-Operator
+
+## 1. Prerequisites
+
+Before deploying the Ceph-CSI-Operator, ensure the following requirements are met:
+
+- A Kubernetes cluster ([supported version](https://kubernetes.io/releases/) recommended)
+- Ceph cluster ([supported version](https://docs.ceph.com/en/latest/releases/) recommended)
```
obnoxxx, line 19: Might be nice to mention the rook project here as a convenient means of creating a Ceph cluster in k8s.

docs/quick-start.md
```diff
@@ -0,0 +1,160 @@
...
+```console
+echo '
+apiVersion: csi.ceph.io/v1alpha1
+kind: CephConnection
+metadata:
+  name: ceph-connection
+  namespace: ceph-csi-operator-system
+spec:
+  monitors:
+  - 10.98.44.171:6789
```
obnoxxx, line 107: It might be useful to explain how to get the requires IP address and port, e. g. when using rook to set up Ceph.

obnoxxx (changes requested):
two comments  inline. Not strictly change requests really, rather suggestions/thoughts.

docs/quick-start.md
Madhu-1: I have kept the options open for deploying Rook and its up to the user to use their own way to deploy the ceph cluster, that why i have mentioned that a supported version of ceph need to be installed.

docs/quick-start.md
Madhu-1: As the ceph cluster is configured I assume that the user knows the basic monitor information to configure anything. that is the reason not pointing to any specific commands as it depends on the way of the ceph cluster deployment.

obnoxxx (approved):
LGTM

Madhu-1:
@nb-ohad PTAL

docs/quick-start.md
nb-ohad: @obnoxxx I agree with @Madhu-1 
Getting the IP address can be different based on how Ceph was deployed (rook or other means), we should not be based toward a specific deployment type and the assumption is that Ceph is already deployed.

nb-ohad (approved):

Issue #78, opened 2024-08-02
Fix CR not found after delete operation


Madhu-1:
I see a pattern of loading operands across all controllers, however an event is always being assumed as the resource will exist but for deletion it may not and this could potentially result in not ending reconciles, isn't it?

_Originally posted by @leelavg in https://github.com/ceph/ceph-csi-operator/pull/48#discussion_r1699884500_

PR #151, opened 2024-10-04
do not always error out while checking for existence of resources

leelavg:
controllers also get reconciles for deletion events and it may happen that resource at api server be deleted first before it being purged from cache and during that time if we error out the reconciliation will never be successful in standard conditions.

fixes: #78

### DIFF
--- a/internal/controller/clientprofile_controller.go
+++ b/internal/controller/clientprofile_controller.go
@@ -130,6 +130,10 @@ func (r *ClientProfileReconciler) Reconcile(ctx context.Context, req ctrl.Reques
 
 func (r *ClientProfileReconcile) reconcile() error {
 	if err := r.loadAndValidate(); err != nil {
+		if utils.IsNotFoundWithName(err, r.clientProfile.Name) {
+			r.log.Info("Client profile resource does not exists anymore, skipping reconcile")
+			return nil
+		}
 		return err
 	}
 
--- a/internal/controller/driver_controller.go
+++ b/internal/controller/driver_controller.go
@@ -226,6 +226,10 @@ func (r *DriverReconciler) Reconcile(ctx context.Context, req ctrl.Request) (ctr
 func (r *driverReconcile) reconcile() error {
 	// Load the driver desired state based on driver resource, operator config resource and default values.
 	if err := r.LoadAndValidateDesiredState(); err != nil {
+		if utils.IsNotFoundWithName(err, r.driver.Name) {
+			r.log.Info("Driver resource does not exist anymore, skipping reconcile")
+			return nil
+		}
 		return err
 	}
 
--- a/internal/utils/meta.go
+++ b/internal/utils/meta.go
@@ -17,6 +17,9 @@ limitations under the License.
 package utils
 
 import (
+	"errors"
+
+	k8serrors "k8s.io/apimachinery/pkg/api/errors"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/runtime"
 	ctrlutil "sigs.k8s.io/controller-runtime/pkg/controller/controllerutil"
@@ -61,4 +64,14 @@ func ToggleOwnerReference(on bool, obj, owner metav1.Object, scheme *runtime.Sch
 		return err == nil, err
 	}
 	return false, nil
+}
+
+func IsNotFoundWithName(err error, name string) bool {
+	if !k8serrors.IsNotFound(err) {
+		return false
+	}
+	status, ok := err.(k8serrors.APIStatus)
+	return (ok || errors.As(err, &status)) &&
+		status.Status().Details != nil &&
+		status.Status().Details.Name == name
 }


leelavg:
I'm a bit divided b/n where the handling should happen, like, err returned from the load function can't always be same as returned by api server.

To be moved inside the function it needs to copied at a couple of places, pls review.

internal/controller/clientprofile_controller.go
```diff
@@ -130,6 +130,10 @@ func (r *ClientProfileReconciler) Reconcile(ctx context.Context, req ctrl.Reques
 
 func (r *ClientProfileReconcile) reconcile() error {
 	if err := r.loadAndValidate(); err != nil {
+		if client.IgnoreNotFound(err) == nil {
```
nb-ohad, line 133: I think what you want is to compare to a Not Found error, not to negate a negation of a Not Found error
```suggestion
		if k8serrors.IsNotFound(err) {
```

internal/controller/driver_controller.go
```diff
@@ -226,6 +226,10 @@ func (r *DriverReconciler) Reconcile(ctx context.Context, req ctrl.Request) (ctr
 func (r *driverReconcile) reconcile() error {
 	// Load the driver desired state based on driver resource, operator config resource and default values.
 	if err := r.LoadAndValidateDesiredState(); err != nil {
+		if client.IgnoreNotFound(err) == nil {
```
nb-ohad, line 229: Same as last comment
```suggestion
		if k8serrors.IsNotFound(err) {
```

internal/controller/clientprofile_controller.go
```diff
@@ -130,6 +130,10 @@ func (r *ClientProfileReconciler) Reconcile(ctx context.Context, req ctrl.Reques
...
+			r.log.Info("Resources for reconciling clientprofile is not found. Object(s) might have been deleted.")
```
nb-ohad, line 134: Let's be more specific in the log message:
```suggestion
			r.log.Info("Client profile resource does not exists anymore, skipping reconcile")
```

internal/controller/driver_controller.go
```diff
@@ -226,6 +226,10 @@ func (r *DriverReconciler) Reconcile(ctx context.Context, req ctrl.Request) (ctr
...
+			r.log.Info("Resources for reconciling driver is not found. Object(s) might have been deleted.")
```
nb-ohad, line 230: Let's be more specific in the log message:
```suggestion
			r.log.Info("Driver resource does not exist anymore, skipping reconcile.")
```

internal/controller/driver_controller.go
```diff
@@ -296,7 +300,7 @@ func (r *driverReconcile) LoadAndValidateDesiredState() error {
 	opConfig := csiv1a1.OperatorConfig{}
 	opConfig.Name = operatorConfigName
 	opConfig.Namespace = operatorNamespace
-	if err := r.Get(r.ctx, client.ObjectKeyFromObject(&opConfig), &opConfig); client.IgnoreNotFound(err) != nil {
+	if err := r.Get(r.ctx, client.ObjectKeyFromObject(&opConfig), &opConfig); err != nil {
```
nb-ohad, line 303: This change does not make sense, the operator config is optional. We need to reconcile even if it does not exists

nb-ohad (changes requested):

internal/controller/clientprofile_controller.go
leelavg: done.

internal/controller/driver_controller.go
leelavg: done.

internal/controller/clientprofile_controller.go
leelavg: done.

internal/controller/driver_controller.go
leelavg: done.

internal/controller/driver_controller.go
```diff
@@ -226,6 +226,10 @@ func (r *DriverReconciler) Reconcile(ctx context.Context, req ctrl.Request) (ctr
 func (r *driverReconcile) reconcile() error {
 	// Load the driver desired state based on driver resource, operator config resource and default values.
 	if err := r.LoadAndValidateDesiredState(); err != nil {
+		if k8serrors.IsNotFound(err) {
+			r.log.Info("Driver resource does not exist anymore, skipping reconcile")
+			return nil
+		}
```
Madhu-1, line 232: I think this is should not be here because it is returning nil for all the not found errors, like if the imageset configmap not found we will get here and we return nil to stop the reconcile which is not correct.

nb-ohad: I agree.
The inner function needs to return a different custom error so the top-most function can make a decision based on that concrete error

leelavg: agreed, done.

internal/controller/clientprofile_controller.go
```diff
@@ -130,6 +131,10 @@ func (r *ClientProfileReconciler) Reconcile(ctx context.Context, req ctrl.Reques
 
 func (r *ClientProfileReconcile) reconcile() error {
 	if err := r.loadAndValidate(); err != nil {
+		if k8serrors.IsNotFound(err) && r.clientProfile.UID == "" {
```
nb-ohad, line 134: @leelavg wouldn't it make more sense to investigate the error object to see the target of the not found error?

internal/controller/driver_controller.go
```diff
@@ -226,6 +226,10 @@ func (r *DriverReconciler) Reconcile(ctx context.Context, req ctrl.Request) (ctr
 func (r *driverReconcile) reconcile() error {
 	// Load the driver desired state based on driver resource, operator config resource and default values.
 	if err := r.LoadAndValidateDesiredState(); err != nil {
+		if k8serrors.IsNotFound(err) && r.driver.UID == "" {
```
nb-ohad, line 229: Same as last comment

nb-ohad (changes requested):

internal/controller/driver_controller.go
leelavg: done.

internal/utils/meta.go
```diff
@@ -62,3 +65,13 @@ func ToggleOwnerReference(on bool, obj, owner metav1.Object, scheme *runtime.Sch
 	}
 	return false, nil
 }
+
+func IsNotFoundWithName(err error, name string) bool {
+	if !k8serrors.IsNotFound(err) {
+		return false
+	}
+	status, ok := err.(k8serrors.APIStatus)
+	return (ok || errors.As(err, &status)) &&
```
nb-ohad, line 74: @leelavg Why the two different approaches? can we always use `errors.As`?

nb-ohad (changes requested):

internal/utils/meta.go
leelavg: there is a possibility that APIStatus could be wrapped and so the interface checks fails but unwrapping it can be successful which provides the `Status()` function satisfying this interface and query the error details.

another thing to note here is, API could just return httpStatusNotFound without setting details of the resource in which case this function fails to work as intended, based on the review I could return true for both `if status not found and resource is intended object`, `if status is not found but resource isn't set for us to verify`.

nb-ohad: We don't need to complicate this so much, we know exactly the function that produced the error and the potential error space coming from it, Also we are looking for a subset of a specific error code that comes from a trusted source (the k8s client). With all of this in mind, you don't need to write code that is so defensive

leelavg:
will re-raise if required.

PR #165, opened 2024-11-04
api: add radosNamespace field to CephFsConfigSpec

iPraveenParihar:

Ceph-CSI now supports storing CephFS OMAP data in a
specified namespace defined in CephFS.radosNamespace
field in ceph-csi-config ConfigMap.
ref: https://github.com/ceph/ceph-csi/pull/4661.

This commit introduces the same by adding the radosNamespace field
to the ClientProfile.ClientProfileSpec.CephFsConfigSpec API.

**Checklist:**

* [x] **Commit Message Formatting**: Commit titles and messages follow guidelines in the [developer guide](https://github.com/ceph/ceph-csi-operator/blob/devel/docs/development-guide.md#commit-messages).
* [x] Reviewed the developer guide on [Submitting a Pull Request](https://github.com/ceph/ceph-csi-operator/blob/devel/docs/development-guide.md#development-workflow)
* [ ] [Pending release notes](https://github.com/ceph/ceph-csi-operator/blob/devel/PendingReleaseNotes.md) updated with breaking and/or notable changes for the next major release.


api/v1alpha1/clientprofile_types.go
```diff
@@ -31,6 +31,9 @@ type CephFsConfigSpec struct {
 
 	//+kubebuilder:validation:Optional
 	FuseMountOptions map[string]string `json:"fuseMountOptions,omitempty"`
+
+	//+kubebuilder:validation:Optional
+	RadosNamespace string `json:"radosNamespace,omitempty"`
```
leelavg, line 36: in the linked PR, it was mentioned this field should be added to newer deployments only, should we support that or will there be any way to migrate the older way to use rns for omap data?

iPraveenParihar: > in the linked PR, it was mentioned this field should be added to newer deployments only, should we support that 

```
# NOTE: Make sure you don't add radosNamespace option to a currently in use
# configuration as it will cause issues.
```

Yes, the radosNamespace field should only be included in new clientProfile configuration to avoid potential issues. 
Users should note that they don't add the radosNamespace option to an existing clientProfile configuration, as it may disrupt current operations.
Since Ceph-CSI supports specifying a radosNamespace for CephFS volumes, the Ceph-CSI operator should also support it.  Thoughts @Madhu-1?

> will there be any way to migrate the older way to use rns for omap data?

At this time, no migration design exists for migrating OMAP data between namespaces.

Madhu-1: No migration is supported from cephcsi side. this field should be immutable and should not be changed upon creation

api/v1alpha1/clientprofile_types.go
```diff
@@ -31,10 +31,15 @@ type CephFsConfigSpec struct {
 
 	//+kubebuilder:validation:Optional
 	FuseMountOptions map[string]string `json:"fuseMountOptions,omitempty"`
+
+	//+kubebuilder:validation:XValidation:rule="self == oldSelf",message="radosNamespace is immutable"
+	//+kubebuilder:validation:Optional
+	RadosNamespace string `json:"radosNamespace,omitempty"`
```
nb-ohad, line 37: The usage of a Rados namespace for CephFs is sepcificly for OMAP data. The current proposed name for this field does not reflects that. I would suggest the following instead:
```suggestion
	OmapNamespace string `json:"OmapNamespace,omitempty"`
```

api/v1alpha1/clientprofile_types.go
```diff
@@ -31,10 +31,15 @@ type CephFsConfigSpec struct {
 
 	//+kubebuilder:validation:Optional
 	FuseMountOptions map[string]string `json:"fuseMountOptions,omitempty"`
+
+	//+kubebuilder:validation:XValidation:rule="self == oldSelf",message="radosNamespace is immutable"
```
nb-ohad, line 35: Why the immutability?

nb-ohad (changes requested):

api/v1alpha1/clientprofile_types.go
iPraveenParihar: https://github.com/ceph/ceph-csi-operator/pull/165#discussion_r1828962621

api/v1alpha1/clientprofile_types.go
iPraveenParihar: OMAP are RADOS objects so the term `radosNamespace` would be more appropriate?

nb-ohad: @iPraveenParihar 
I am not sure I agree with the lack of distinction. You want the user to specify a rados namespace for the location of the omap data for the cephfs usage. Not just a general rados namespace for general use. 

What I am saying is that the name should capture that
Maybe we could agree on `OmapRadosNamespace` as an alternative?

Rakshith-R: It's radosNamespace in cephcsi, let's keep it the same.

Currently it may well be only used for omap usage but it might change in the future.

nb-ohad: User-facing API design is essential for me,  there is already past acknowledgment that the Ceph CSI project did not select good names for some of its front-facing configurations. Bringing them as an example does not make a good case. 

Regarding the other part of the comment:
1.  we are talking about CephFS, not RBD. RDS is the internal mechanism we use to isolate the omap data. It does not capture the value/intent  of using the feature/nob
2. Using the same field name as we have inside the RBD config section is misleading because in each section the nob had completely different semantics and effects. 
3. Using RadosNamesapce as the nob name will collide with any future nobs that we might need to allow configuration of additional RDS for other metadata the cephcsi or chepfs will allow us to use. We want future proofing against such a case. 


Given the above 3 points, we should go with something like OmapRadosNamaspece which captures both the intent of the nob and the entity type this nob is referring to

Rakshith-R: > User-facing API design is essential for me, there is already past acknowledgment that the Ceph CSI project did not select good names for some of its front-facing configurations. Bringing them as an example does not make a good case.

I completely disagree with the statement `Ceph CSI project did not select good names`. 
You can always open issues at cephcsi to suggest renaming them.

Changing such conventions in above layers only causes confusion. 

> 
> Regarding the other part of the comment:
> 
> 1. we are talking about CephFS, not RBD. RDS is the internal mechanism we use to isolate the omap data. It does not capture the value/intent  of using the feature/nob
> 2. Using the same field name as we have inside the RBD config section is misleading because in each section the nob had completely different semantics and effects.
> 3. Using RadosNamesapce as the nob name will collide with any future nobs that we might need to allow configuration of additional RDS for other metadata the cephcsi or chepfs will allow us to use. We want future proofing against such a case.
> 
> Given the above 3 points, we should go with something like OmapRadosNamaspece which captures both the intent of the nob and the entity type this nob is referring to

RadosNamespace is the key we are using for cephfs section in cephcsi, we should keep it the same in cephcsi Operator. It implies cephcsi will use this radosNS for operations, right now it maybe limited to just omap but might change in the future.

If you would like more specific naming, please open issue in cephcsi, we'll change it there, then in Operator. I would prefer this approach instead of introducing new names abruptly in above layers.

Future additional cases can use another specific key.

iPraveenParihar: @Madhu-1, can you please provide your inputs here?

black-dragon74:
We need this for RHSTOR-6421

cc: @nb-ohad @Madhu-1

api/v1alpha1/clientprofile_types.go
nb-ohad: Ok in that case, can we at least call it `MetadataRadosNamespace`?

api/v1alpha1/clientprofile_types.go
```diff
@@ -31,10 +31,15 @@ type CephFsConfigSpec struct {
...
+	//+kubebuilder:validation:Optional
+	RadosNamespace string `json:"radosNamespace,omitempty"`
 }
 
 // RbdConfigSpec defines the desired RBD configuration
 type RbdConfigSpec struct {
+	//+kubebuilder:validation:XValidation:rule="self == oldSelf",message="radosNamespace is immutable"
```
nb-ohad, line 42: Lets not do that. We might want to use changing this name when we unfortunately need to migrate data between rados namespaces

api/v1alpha1/clientprofile_types.go
nb-ohad, line 37: When it comes to rados namespaces, the empty string is a valid namespace name and is used for the implicit namespace. In this case an optional value should be backed by a pointer to a string to differentiate the implicit namespace from non setting a namespace at all that will default to the behavior which might or might not be using the implicit namespace

nb-ohad (changes requested):

api/v1alpha1/clientprofile_types.go
Madhu-1: cephcsi doesnt support empty namespace, am fine with having it as a pointer

Madhu-1: @nb-ohad IMHO using RadosNamespace is fine here i dont see a reason to call it MetadataRadosNamespace. Lets not dictate the current use of the field as its limited for the metadata

nb-ohad: Decided offline to use the name `radosNamespace`

vendor/github.com/ceph/ceph-csi-operator/api/v1alpha1/clientprofile_types.go
```diff
@@ -31,10 +31,15 @@ type CephFsConfigSpec struct {
 
 	//+kubebuilder:validation:Optional
 	FuseMountOptions map[string]string `json:"fuseMountOptions,omitempty"`
+
+	//+kubebuilder:validation:XValidation:rule="self == oldSelf",message="radosNamespace is immutable"
+	//+kubebuilder:validation:Optional
+	RadosNamespace *string `json:"radosNamespace"`
```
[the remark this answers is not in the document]
iPraveenParihar: I don't see any misalignment between the original and the generated file. Could you please highlight the misaligned content? Thanks!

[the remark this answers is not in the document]
Madhu-1: Can you please rebase your PR?

Madhu-1 (dismissed):

internal/controller/clientprofile_controller.go
```diff
@@ -313,6 +314,9 @@ func composeCsiClusterInfoRecord(clientProfile *csiv1a1.ClientProfile, cephConn
 	record.Monitors = cephConn.Spec.Monitors
 	if cephFs := clientProfile.Spec.CephFs; cephFs != nil {
 		record.CephFs.SubvolumeGroup = cephFs.SubVolumeGroup
+		if cephfsRadosNamespace := cephFs.RadosNamespace; cephfsRadosNamespace != nil {
+			record.CephFs.RadosNamespace = *cephfsRadosNamespace
+		}
```
nb-ohad, line 319:
```suggestion
		record.CephFs.RadosNamespace = cmp.Or(cephfsRadosNamespace, "") 
```

docs/design/operator.md
```diff
@@ -349,8 +349,9 @@ spec:
     subvolumeGroup: csi
     kernelMountOptions: readdir_max_bytes=1048576,norbytes
     fuseMountOptions: debug
+    radosNamespace: rados-test-cephfs
   rbd:
-    radosNamespace: rados-test
+    radosNamespace: rados-test-rbd
```
nb-ohad, line 354: There is not equivalence between the RBD RADOS NS and the RADOS NS for CephFS metadata.
With that in mind, please rename this back to: 
```suggestion
    radosNamespace: rados-test
```

nb-ohad (changes requested):

internal/controller/clientprofile_controller.go
```diff
@@ -313,6 +315,9 @@ func composeCsiClusterInfoRecord(clientProfile *csiv1a1.ClientProfile, cephConn
 	record.Monitors = cephConn.Spec.Monitors
 	if cephFs := clientProfile.Spec.CephFs; cephFs != nil {
 		record.CephFs.SubvolumeGroup = cephFs.SubVolumeGroup
+		if cephfsRadosNamespace := cephFs.RadosNamespace; cephfsRadosNamespace != nil {
+			record.CephFs.RadosNamespace = cmp.Or(*cephfsRadosNamespace, "")
+		}
```
nb-ohad, line 320: Why do we need the if statement if we use `cmp.Or`?

```suggestion
		record.CephFs.RadosNamespace = cmp.Or(cephFs.RadosNamespace, "")
```

nb-ohad (changes requested):

internal/controller/clientprofile_controller.go
iPraveenParihar: Since we made `cephFs.RadosNamespace` a pointer _https://github.com/ceph/ceph-csi-operator/pull/165#discussion_r1983185948_, the field will be `nil` if it is absent. However, `cmp.Or` does not handle `nil` checks.

Am I missing something?

nb-ohad: Let try something different
```suggestion
		record.CephFs.RadosNamespace = ptr.Deref(cephfsRadosNamespace, "")
```

iPraveenParihar: Thanks!

PR #172, opened 2024-11-13
csiaddons: add support for TLS

bipuladh:
# Description #
This PR introduces a controller to approve CSR requests intended for Ceph CSI TLS communications.

## Context 

To have a secure gRPC communication between the CSI add-ons sidecar and CSI manager we need to create certificates that can be verified by a CA. We use the k8s CSR resource to sign our certificates. Related to: https://github.com/csi-addons/kubernetes-csi-addons/pull/692

## Is the change backward compatible? 
No

## Are there concerns around backward compatibility?

We will move ahead with disabled by default approach so that it doesn't introduce breakages.

For example:

**Checklist:**

  guidelines in the [developer
* [ ] [Pending release


internal/controller/csr_controller.go
```diff
@@ -0,0 +1,132 @@
+/*
```
bipuladh, line 1: As discussed with the team the CSI operator should not be responsible for generation of certificates. ODF operators can generate these certificates.

leelavg:
pls excuse I partly understood what the code is doing but don't get the "why". Why should csi-op behave like a trimmed down CertificateAuthority? A CA has much more functionalities like rotation, revocation to name a few in addition to a lot of options in x509 cert.

bipuladh:
This PR is not applicable anymore because of the different approach we took for this feature request.

PR #178, opened 2024-11-20
Make it easy to run a single test

raghavendra-talur:
This PR moves the config init code to the main function from the global init. Doing so, it allows for such config init to be skipped for tests or other scenarios.

In the second commit, we set the os environment variable for the operator namespace in the BeforeSuite() function if it is not already set. This allows users to run a specific test by using the ginkgo cli or by using the test decorators in vscode.

For example:

```
$ ginkgo -focus "ClientProfile Controller" -r
[1732083376] Controller Suite - 1/3 specs SS• SUCCESS! 5.797831s PASS
  Starting ceph-csi-operator suite
[1732083376] e2e suite - 0/1 specs S SUCCESS! 59.25µs PASS

Ginkgo ran 2 suites in 10.605577042s
Test Suite Passed
```

Screenshot 2024-11-20 at 1 24 33 AM

Earlier, the configuration was automatically inited. Now, it is being done manually in the main function. 

Yes

No

**Checklist:**

  guidelines in the [developer
* [x] [Pending release
* [x] Unit tests have been added, if necessary.
* [x] Integration tests have been added, if necessary.

### DIFF
--- a/internal/controller/defaults.go
+++ b/internal/controller/defaults.go
@@ -24,7 +24,6 @@ import (
 	"k8s.io/utils/ptr"
 
 	csiv1a1 "github.com/ceph/ceph-csi-operator/api/v1alpha1"
-	"github.com/ceph/ceph-csi-operator/internal/utils"
 )
 
 var imageDefaults = map[string]string{
@@ -65,25 +64,26 @@ var defaultDeploymentStrategy = appsv1.DeploymentStrategy{
 	},
 }
 
-var operatorNamespace = utils.Call(func() string {
-	namespace := os.Getenv("OPERATOR_NAMESPACE")
-	if namespace == "" {
+var (
+	operatorNamespace    string
+	operatorConfigName   string
+	serviceAccountPrefix string
+)
+
+func InitConfig() {
+	if operatorNamespace = os.Getenv("OPERATOR_NAMESPACE"); operatorNamespace == "" {
 		panic("Required OPERATOR_NAMESPACE environment variable is either missing or empty")
 	}
-	return namespace
-})
 
-var operatorConfigName = utils.Call(func() string {
-	name, ok := os.LookupEnv("OPERATOR_CONFIG_NAME")
-	if ok {
-		if name == "" {
+	serviceAccountPrefix = os.Getenv("CSI_SERVICE_ACCOUNT_PREFIX")
+
+	envOperatorConfigName, set := os.LookupEnv("OPERATOR_CONFIG_NAME")
+	if set {
+		if envOperatorConfigName == "" {
 			panic("OPERATOR_CONFIG_NAME exists but empty")
 		}
-		return name
+		operatorConfigName = envOperatorConfigName
+	} else {
+		operatorConfigName = "ceph-csi-operator-config"
 	}
-	return "ceph-csi-operator-config"
-})
-
-var serviceAccountPrefix = utils.Call(func() string {
-	return os.Getenv("CSI_SERVICE_ACCOUNT_PREFIX")
-})
+}
--- a/internal/controller/suite_test.go
+++ b/internal/controller/suite_test.go
@@ -18,6 +18,7 @@ package controller
 
 import (
 	"fmt"
+	"os"
 	"path/filepath"
 	"runtime"
 	"testing"
@@ -39,9 +40,12 @@ import (
 // These tests use Ginkgo (BDD-style Go testing framework). Refer to
 // http://onsi.github.io/ginkgo/ to learn more about Ginkgo.
 
-var cfg *rest.Config
-var k8sClient client.Client
-var testEnv *envtest.Environment
+var (
+	cfg                   *rest.Config
+	k8sClient             client.Client
+	testEnv               *envtest.Environment
+	operatorNSUnsetNeeded bool
+)
 
 func TestControllers(t *testing.T) {
 	RegisterFailHandler(Fail)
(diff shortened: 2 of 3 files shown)


leelavg:
Isn't only second commit is enough for the required behavior, assuming the behavior is to have operator namespace to be set? First commit just delays the initialization w/o any functionality change, isn't it?

raghavendra-talur:
> Isn't only second commit is enough for the required behavior, assuming the behavior is to have operator namespace to be set? First commit just delays the initialization w/o any functionality change, isn't it?

Good question. If I run the test with just the commit that sets the env var then I see this error:

```
$ ginkgo -focus "ClientProfile Controller" -r
panic: Required OPERATOR_NAMESPACE environment variable is either missing or empty

goroutine 1 [running]:
github.com/ceph/ceph-csi-operator/internal/controller.init.func1(...)
        /Users/rtalur/src/github.com/raghavendra-talur/ceph-csi-operator/internal/controller/defaults.go:71
github.com/ceph/ceph-csi-operator/internal/utils.Call[...](...)
        /Users/rtalur/src/github.com/raghavendra-talur/ceph-csi-operator/internal/utils/core.go:121
github.com/ceph/ceph-csi-operator/internal/controller.init()
        /Users/rtalur/src/github.com/raghavendra-talur/ceph-csi-operator/internal/controller/defaults.go:68 +0x3c8

Ginkgo ran 2 suites in 6.842060084s

There were failures detected in the following suites:
  controller ./internal/controller

Test Suite Failed
```

The reason is that these variables were being assigned at the global scope by calling a function.

```
var operatorNamespace = utils.Call(func() string {
        namespace := os.Getenv("OPERATOR_NAMESPACE")
        if namespace == "" {
                panic("Required OPERATOR_NAMESPACE environment variable is either missing or empty")
        }
        return namespace
})
```

This makes these function calls part of the Go Init() process. The setting of env var in suite_test.go would be run *after* the Init() process but that would be too late as this function panics on not finding the env var set.

internal/controller/suite_test.go
```diff
@@ -66,6 +67,11 @@ var _ = BeforeSuite(func() {
 			fmt.Sprintf("1.29.0-%s-%s", runtime.GOOS, runtime.GOARCH)),
 	}
 
+	_, set := os.LookupEnv("OPERATOR_NAMESPACE")
+	if !set {
+		Expect(os.Setenv("OPERATOR_NAMESPACE", "ceph-csi-operator-system")).To(Succeed())
```
Madhu-1, line 72: in the AfterSuit lets unset this env if we are setting it

raghavendra-talur: Done

internal/controller/defaults.go
```diff
@@ -65,25 +64,24 @@ var defaultDeploymentStrategy = appsv1.DeploymentStrategy{
...
-	namespace := os.Getenv("OPERATOR_NAMESPACE")
-	if namespace == "" {
+var (
+	operatorNamespace    string
+	operatorConfigName   string
+	serviceAccountPrefix string
+)
+
+func InitConfig() {
+	if operatorNamespace := os.Getenv("OPERATOR_NAMESPACE"); operatorNamespace == "" {
```
leelavg, line 74: Isn't the walrus shawdowing global operatorNamespace var here?

leelavg:
Possible to hold off the merge till tomorrow? pls excuse I missed checking this today.

The approach isn't sitting well w/ me for a couple of reasons, BTW open to be corrected.
1. We are creating an indirect dependency on the pkg that something MUST be done in addition to importing it for controllers to work
2. Amending a fundamental requirement to be able to run test, kinda inverted.

raghavendra-talur:
> Possible to hold off the merge till tomorrow? pls excuse I missed checking this today.
> 
> The approach isn't sitting well w/ me for a couple of reasons, BTW open to be corrected.
> 1. We are creating an indirect dependency on the pkg that something MUST be done in addition to importing it for controllers to work
> 2. Amending a fundamental requirement to be able to run test, kinda inverted.

Yes, I am in no hurry to get this merged. I will reply with examples to answer the above questions.

leelavg:
Looked again, the fix introduces a bug in both test & runtime environments which I already left a comment.

I couldn't find any simpler alternatives, I don't use the feature (vscode & ginkgo) which requires this fix and leave the decision to maintainers.

BTW, see if `t.Setenv` helps in anyway which removes explicit cleanup.

internal/controller/defaults.go
raghavendra-talur: Fixed, thanks!

raghavendra-talur:
> The approach isn't sitting well w/ me for a couple of reasons, BTW open to be corrected.
> 
> 1. We are creating an indirect dependency on the pkg that something MUST be done in addition to importing it for controllers to work


This is not true. Just importing the controllers package in main.go wasn't enough even before. The SetupWithManager() has to be invoked for every controller in the package, which are 3 at this time, ClientProfileMappingReconciler, ClientProfileReconciler, and DriverReconciler. 

> 2. Amending a fundamental requirement to be able to run test, kinda inverted.
I don't understand which requirement is amended here?

raghavendra-talur:
Ok, the tests have passed but now I agree with @leelavg that it is starting to look like a big change for a project which I am not the maintainer.

@nb-ohad @Madhu-1 I will leave it up to you to merge or close the PR.

Ginkgo and VSCode both provide users a way to specify a env var file and that is one alternative option for the users.

Madhu-1:
> Ginkgo and VSCode both provide users a way to specify an env var file and that is one alternative option for the users.

@raghavendra-talur we are reworking on the new framework for the testing, Lets keep this PR open for sometime and we can see if we can get this in or we need any additional changes later on.

Issue #184, opened 2024-12-04
High memory usage due to watching all namespaces


leelavg:
```
---------------------------------------- when memory request is 64Mi and limit is removed
sh-5.1$ echo =====; echo start at $(date --iso-8601=seconds), usage is [$(awk '{print $1/1024/1024/1024 " Gi"}' /sys/fs/cgroup/memory/memory.usage_in_bytes)], limit is [$(awk '{print $1/1024/1024/1024 " Gi"}' /sys/fs/cgroup/memory/memory.limit_in_bytes)] && (/manager > /tmp/manager.log 2>&1 &) && prev=$(date +%s); while true; do now=$(date +%s); echo -n $((now-prev))s:; awk '{print $1/1024/1024/1024 " Gi"}' /sys/fs/cgroup/memory/memory.usage_in_bytes; sleep 2; done;
=====
start at 2024-12-04T05:09:47+00:00, usage is [0.00228882 Gi], limit is [8.58993e+09 Gi]
0s:0.0105362 Gi
2s:0.0860825 Gi
4s:0.149323 Gi
6s:0.241344 Gi
8s:0.272717 Gi
10s:0.363647 Gi
12s:0.587654 Gi
15s:0.665798 Gi
17s:0.714954 Gi
19s:0.813492 Gi
21s:0.812382 Gi
23s:1.23141 Gi
25s:1.46246 Gi <<< process got stabilized at a peak usage of ~1.5Gi
27s:1.46161 Gi
29s:1.46153 Gi
31s:1.46188 Gi
33s:1.46174 Gi
35s:1.46168 Gi
37s:1.46179 Gi
39s:1.46175 Gi
41s:1.46176 Gi
^C

---------------------------------------- when memory request is 64Mi and limit is 128Mi (set during build time)
sh-5.1$ echo =====; echo start at $(date --iso-8601=seconds), usage is [$(awk '{print $1/1024/1024/1024 " Gi"}' /sys/fs/cgroup/memory/memory.usage_in_bytes)], limit is [$(awk '{print $1/1024/1024/1024 " Gi"}' /sys/fs/cgroup/memory/memory.limit_in_bytes)] && (/manager > /tmp/manager.log 2>&1 &) && prev=$(date +%s); while true; do now=$(date +%s); echo -n $((now-prev))s:; awk '{print $1/1024/1024/1024 " Gi"}' /sys/fs/cgroup/memory/memory.usage_in_bytes; sleep 2; done;
=====
[...]

PR #186, opened 2024-12-11
restrict manager resource cache based on namespaces from environment

leelavg:
always cache resources from operator deployed namespace and provider an option to user for caching resources from other namespaces via `WATCH_NAMESPACE` env var with a comma separated namespace values.

fixes: #184

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -7,6 +7,8 @@ IMAGE_NAME ?= ceph-csi-operator
 # Allow customization of the name prefix and/or namespace
 NAME_PREFIX ?= ceph-csi-operator-
 NAMESPACE ?= $(NAME_PREFIX)system
+# A comma separated list of namespaces for operator to cache objects from
+WATCH_NAMESPACE ?= ""
 
 IMG ?= $(IMAGE_REGISTRY)/$(REGISTRY_NAMESPACE)/$(IMAGE_NAME):$(IMAGE_TAG)
 
@@ -47,6 +49,11 @@ patches:
       value:
         name: CSI_SERVICE_ACCOUNT_PREFIX
         value: $(NAME_PREFIX)
+    - op: add
+      path: /spec/template/spec/containers/1/env/-
+      value:
+        name: WATCH_NAMESPACE
+        value: $(WATCH_NAMESPACE)
   target:
     kind: Deployment
     name: controller-manager
--- a/cmd/main.go
+++ b/cmd/main.go
@@ -19,7 +19,9 @@ package main
 import (
 	"crypto/tls"
 	"flag"
+	"fmt"
 	"os"
+	"strings"
 
 	// Import all Kubernetes client auth plugins (e.g. Azure, GCP, OIDC, etc.)
 	// to ensure that exec-entrypoint and run can make use of them.
@@ -29,6 +31,7 @@ import (
 	utilruntime "k8s.io/apimachinery/pkg/util/runtime"
 	clientgoscheme "k8s.io/client-go/kubernetes/scheme"
 	ctrl "sigs.k8s.io/controller-runtime"
+	"sigs.k8s.io/controller-runtime/pkg/cache"
 	"sigs.k8s.io/controller-runtime/pkg/healthz"
 	"sigs.k8s.io/controller-runtime/pkg/log/zap"
 	metricsserver "sigs.k8s.io/controller-runtime/pkg/metrics/server"
@@ -36,6 +39,7 @@ import (
 
 	csiv1alpha1 "github.com/ceph/ceph-csi-operator/api/v1alpha1"
 	"github.com/ceph/ceph-csi-operator/internal/controller"
+	"github.com/ceph/ceph-csi-operator/internal/utils"
 	//+kubebuilder:scaffold:imports
 )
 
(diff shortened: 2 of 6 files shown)


leelavg:
tested, peak usage was 50 Mi for startup down from ~ 1.5 Gi on a semi-prod cluster.

Madhu-1:
The namespace(s) name needs to be configurable as this operator is developed to work with multiple namespaces

https://sdk.operatorframework.io/docs/building-operators/golang/operator-scope/#configuring-watch-namespaces-dynamically is the way we need to look to fix this one.

leelavg:
Yes, I want to refactor when we actually need it and so reused operator namespace.

leelavg:
> The namespace(s) name needs to be configurable as this operator is developed to work with multiple namespaces
> 
> https://sdk.operatorframework.io/docs/building-operators/golang/operator-scope/#configuring-watch-namespaces-dynamically is the way we need to look to fix this one.
- not working, need to go through controller-runtime code on what changed.

leelavg:
done.

cmd/main.go
```diff
@@ -160,3 +181,25 @@ func main() {
...
+
+	ns := os.Getenv(watchNamespaceEnvVar)
+	if ns == "" {
+		return "", fmt.Errorf("%s must be set", watchNamespaceEnvVar)
+	}
+	return ns, nil
+}
+
+// getOperatorNamespace returns the Namespace the operator is running
+func getOperatorNamespace() (string, error) {
```
Madhu-1, line 197: we already have `OperatorNamespace` in utils. can you please use it?

cmd/main.go
```diff
@@ -160,3 +181,25 @@ func main() {
 		os.Exit(1)
 	}
 }
+
+// getWatchNamespace returns the Namespace the operator should be watching for changes
+func getWatchNamespace() (string, error) {
+	var watchNamespaceEnvVar = "WATCH_NAMESPACE"
```
Madhu-1, line 187: can you please change for WATCH_NAMESPACE in makefile and container ENV as we did for serviceaccount_name_prefix https://github.com/ceph/ceph-csi-operator/pull/116?

leelavg: I'm trying not to touch manifests as during backports it's always creating extra work and that should be fixed first.

cmd/main.go
leelavg: Could you pls point, didn't find it. There's something similar in controllers but not usable.

cmd/main.go
Madhu-1: please open a tracker if you are planning to fix it as followupl

cmd/main.go
Madhu-1: sorry i was looking at my local copy where the getOperatorNamespace was exposed, can you please update https://github.com/ceph/ceph-csi-operator/blob/bdd5baa89f31c9f3d3ebfc17edcd3c7768931e55/internal/controller/defaults.go#L68 to use it outside it would duplicate functions

cmd/main.go
```diff
@@ -94,6 +97,23 @@ func main() {
...
 	})
 
+	defaultNamespaces := map[string]cache.Config{}
+	operatorNamespace, err := getOperatorNamespace()
+	if err != nil {
+		setupLog.Error(err, "manager requires namespace to be registered for controllers to reconcile")
+		os.Exit(1)
+	}
+	// ensure we always cache items from operator namespace
+	defaultNamespaces[operatorNamespace] = cache.Config{}
```
Madhu-1, line 107: how will this change work where the operator should be able to manage the csi drivers in all the namespaces?  https://github.com/ceph/ceph-csi-operator/blob/main/docs/design/operator.md#goals-and-guidelines

leelavg: in my tests probably csi driver was already deployed, do you know how rook does it?

leelavg: Rook isn't doing anything explicitly for cluster scoped resources and based on my testing in https://github.com/red-hat-storage/ocs-operator/pull/2911 we do get events for them, ocs-op only watches 2 namespaces.

cmd/main.go
leelavg: as mentioned _in controllers but not usable_, that function serves different purpose, I can only move the env to const.

cmd/main.go
leelavg: retested, it's working w/o any changes
```
> date; k get csidriver  -w
Fri Jan  3 06:26:07 UTC 2025
NAME               ATTACHREQUIRED   PODINFOONMOUNT   STORAGECAPACITY   TOKENREQUESTS   REQUIRESREPUBLISH   MODES        AGE
rbd.csi.ceph.com   true             false            false             <unset>         false               Persistent   64s
rbd.csi.ceph.com   true             false            false             <unset>         false               Persistent   87s
rbd.csi.ceph.com   true             false            false             <unset>         false               Persistent   0s

> date; k delete csidriver rbd.csi.ceph.com; date
Fri Jan  3 06:26:31 UTC 2025
csidriver.storage.k8s.io "rbd.csi.ceph.com" deleted
Fri Jan  3 06:26:31 UTC 2025

Comment (Madhu-1) [reply cmd/main.go:107]: its not for Rook, its for standalone deployment of csi driver

File: cmd/main.go [hunk 4]
```diff
@@ -94,6 +97,23 @@ func main() {
...
+
+	watchNamespace, err := getWatchNamespace()
+	if err != nil {
+		setupLog.Error(err, "manager will only watch for resources in the operator deployed namespace")
```
Madhu-1, line 111: os.Exit(1) needs to be done as its an actual error and the user might expect that things are working for the watch namespace but the operator is silently ignoring it.

cmd/main.go
Madhu-1: I am trying to avoid duplication of the ENV var, so if we change something tomorrow, I don't need to change it in multiple places.

Makefile
```diff
@@ -7,6 +7,8 @@ IMAGE_NAME ?= ceph-csi-operator
 # Allow customization of the name prefix and/or namespace
 NAME_PREFIX ?= ceph-csi-operator-
 NAMESPACE ?= $(NAME_PREFIX)system
+# A comma separated list of namespaces for operator to cache objects from
+WATCH_NAMESPACE ?= ""
```
Madhu-1, line 11: use `NAMESPACE` as default instead of an empty string?

cmd/main.go
leelavg: If the namespace exists we wouldn't get an error and watchnamespace is additional env as we always use operator namespace in the cache.

cmd/main.go
leelavg: No, I meant rook also deploys csidriver even if it is watching single namespace.

Makefile
leelavg: I suppose we can rename this to ADDITIONAL_WATCH_NAMESPACES or always have operator ns in this.

I didn't go with latter as the ns is known by downward api which has different semantics and so left this as user configurable option.

cmd/main.go
leelavg: done.

Madhu-1 (approved):

Issue #202, opened 2025-01-31
Wrong tag is used for image in release branch


subhamkrai:
# Describe the bug #

When looking at the release branch it should point to specific release tag ex v0.2.0 but instead, it has the latest tag https://github.com/ceph/ceph-csi-operator/blob/a0838ff67d5b6fd2b7592bdcaada98a1600480a6/deploy/all-in-one/install.yaml#L15519

# Expected behavior #

Release branch yaml should have specific tag in image instead of latest tag.

PR #205, opened 2025-02-19
add helm charts to the repo

Madhu-1:
use the new helmify helm plugin to generate the helm charts for the repo

helm doc http://book.kubebuilder.io/plugins/available/helm-v1-alpha.html?highlight=helm#helm-plugin-helmv1-alpha

As above is still alpha version and missing the kustomization support and the above one is contributed based on the idea of helmify using helmify for now.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -196,6 +196,14 @@ build-installer: manifests generate kustomize ## Generate a consolidated YAML wi
 	$(KUSTOMIZE) build build > deploy/all-in-one/install.yaml
 	rm -rf build
 
+.PHONY: build-helm-installer
+build-helm-installer: manifests generate kustomize helmify ## Generate helm charts for the operator.
+	mkdir -p build deploy
+	cd build && echo "$$BUILD_INSTALLER_OVERLAY" > kustomization.yaml
+	cd build && $(KUSTOMIZE) edit add resource ../config/default/
+	$(KUSTOMIZE) build build | $(HELMIFY) deploy/charts/ceph-csi-operator
+	rm -rf build
+
 .PHONY: build-multifile-installer
 build-multifile-installer: build-csi-rbac manifests generate kustomize
 	mkdir -p build deploy/multifile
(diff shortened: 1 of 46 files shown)


Madhu-1:
@subhamkrai can you please check if this is enough for helm charts integration

cc @nb-ohad @travisn

deploy/chart/values.yaml
```diff
@@ -0,0 +1,77 @@
...
+      capabilities:
+        drop:
+        - ALL
+      readOnlyRootFilesystem: true
+    env:
+      csiServiceAccountPrefix: ceph-csi-operator-
+      watchNamespace: ""
+    image:
+      repository: quay.io/cephcsi/ceph-csi-operator
+      tag: latest
```
travisn, line 45: We would need a version-specific tag published when a version is released.

deploy/chart/values.yaml
```diff
@@ -0,0 +1,77 @@
...
+    - --upstream=http://127.0.0.1:8080/
+    - --v=0
+    containerSecurityContext:
+      allowPrivilegeEscalation: false
+      capabilities:
+        drop:
+        - ALL
+      readOnlyRootFilesystem: true
+    image:
+      repository: gcr.io/kubebuilder/kube-rbac-proxy
```
travisn, line 20: Is this just for building, or does the helm chart install the rbac proxy container? I would expect the helm chart to only install the csi operator.

deploy/chart/values.yaml
travisn: And how will the csi operator be versioned?

Madhu-1: @travisn Yes this is only for main branch, we need to adjust the script to change the tag based on the release version

deploy/chart/values.yaml
Madhu-1: Today csi-operator container 2 containers one is rbac proxy and another is csi-operator manager, This is auto generated files(helm charts).

deploy/chart/values.yaml
subhamkrai: @Madhu-1 JFYI #202

deploy/chart/values.yaml
subhamkrai: having 2 commits one for auot-generated files and one for actually changes will make easier to review wdyt?

deploy/chart/values.yaml
Madhu-1: @subhamkrai i will take care of it, Currently there is no way to automate it, i have to take care of it manually before releases.

subhamkrai: okay, just wanted to raise here

deploy/chart/values.yaml
Madhu-1: @subhamkrai with this everything is autogenerated except makefile.

nb-ohad (changes requested):

Makefile
```diff
@@ -189,10 +189,11 @@ docker-buildx: ## Build and push docker image for the manager for cross-platform
 	rm Dockerfile.cross
 
 .PHONY: build-installer
-build-installer: manifests generate kustomize ## Generate a consolidated YAML with CRDs and deployment.
+build-installer: manifests generate kustomize helmify ## Generate a consolidated YAML with CRDs, deployment and helm charts.
 	mkdir -p build deploy/all-in-one
 	cd build && echo "$$BUILD_INSTALLER_OVERLAY" > kustomization.yaml
 	cd build && $(KUSTOMIZE) edit add resource ../config/default/
+	$(KUSTOMIZE) build build | $(HELMIFY) deploy/charts/ceph-csi-operator
```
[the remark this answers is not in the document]
Madhu-1: added a new target for chart generation

Issue #43, opened 2024-07-17
CRD apiVersion for stable features running in production should be v1beta1


travisn:
## Describe the feature you'd like to have #

The CSI operator CRDs should be `v1` instead of `v1alpha1` when we expect it to be running in production.

The conversation was started in the csi operator design doc [here](https://github.com/ceph/ceph-csi-operator/pull/1#discussion_r1609012747), but it was marked resolved after some offline conversation without documenting the reasoning. Let's discuss openly on the expectation for v1.

## Who is the end user and what is the use case where this feature will be valuable?
  
Users look at CRD stability for a clue if it is production ready.
We are taking a stable CSI driver and moving the settings to a stable CSI operator where we expect it to quickly be used in production. Production users will expect a stable version, even though it's new. 

Many users may not care about the api version as long as we claim to support it, but some users will look at an alpha version and refuse to run it in production. For Rook upstream I consider it a blocker to use alpha CRDs in production deployments.

If we are anyway supporting it in production, we have to provide full support for all of our customers, for all new and upgraded clusters, with full feature parity for all customized settings. To me, that is the definition of v1, not alpha. Why not just ship with v1 now? Then we don’t have to worry about conversion to v1 later. As suggested, we expect stability from the start if we are putting it in production.

And if we are supporting alpha in production, let's define the blocking criteria for moving it to v1.

travisn:
After some offline discussion, the blocker for declaring stable upstream would be to declare at least `v1beta1`.

Madhu-1:
we have v1 CRD now, Closing it as fixed

PR #216, opened 2025-03-18
api: update api version to v1beta1

black-dragon74:

This patch updates the API version from `v1alpha1` to `v1beta1`

Closes: https://github.com/ceph/ceph-csi-operator/issues/43


Madhu-1:
@black-dragon74 have you tested this change? the ceph-csi-operator should surve both Alpha and Beta API's as we already have clusters for Alpha CRD.

nb-ohad (commented):
I believe that the PR is missing a conversion webhook. Without it k8s will not be able to convert forward and backward between the two version (even if the are identical) 

Have we tested having a v1beta1 CR served as a v1alpha1 CR?

black-dragon74:
> I believe that the PR is missing a conversion webhook.

Closing this one and following it at: #218

PR #219, opened 2025-03-19
chart: Add helm charts for ceph-csi drivers

Madhu-1:
Adding a helm chart to create all the required CR's to have successful 3 csi drivers(cephfs/rbd/nfs)

### DIFF
--- a/.github/workflows/test-chart.yml
+++ b/.github/workflows/test-chart.yml
@@ -7,8 +7,8 @@ permissions:
   contents: read
 
 jobs:
-  test-operator:
-    name: operator chart
+  test-helm:
+    name: helm charts
     runs-on: ubuntu-latest
     steps:
       - name: Clone the code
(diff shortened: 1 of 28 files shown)


Madhu-1:
CI is broken, will test it in CI once its resolved

Madhu-1:
https://github.com/Madhu-1/ceph-csi-operator/actions/runs/13952576847/job/39055692157?pr=1 Test result for helm charts

.github/workflows/test-chart.yml
```diff
@@ -39,25 +39,55 @@ jobs:
         run: |
           make docker-build
 
-      - name: Install Helm
+      - name: Install helm
```
leelavg, line 42: I believe explicit installation isn't required as per https://github.com/actions/runner-images/blob/main/images/ubuntu/Ubuntu2404-Readme.md, I suppose kubectl is also being used from existing env (or installed during minikube)?

Nevertheless, I think it's better to reuse whatever is available if they are in required version range?

Madhu-1: using version explicitly to avoid dependency problem on the base image.

Madhu-1: And also we will test with latest releases from helm

nb-ohad:
LGTM

Madhu-1:
rebased with main

@nb-ohad approval is required to merge the PR.

nb-ohad (approved):

PR #234, opened 2025-04-10
ci: install the tools required for linting

Madhu-1:
Its not possible to get the third party github action to be used automatically in the ceph org, All the actions need to be audited, The currently used one are having problems, Replacing it with our own until we find a fix for using third party actions again.


nb-ohad:
@Madhu-1 Can we please open an issue to track the intent of reintroducing third part actions in the future?

Madhu-1:
> @Madhu-1 Can we please open an issue to track the intent of reintroducing third part actions in the future?

@nb-ohad https://github.com/ceph/ceph-csi-operator/issues/237 here it is

nb-ohad (approved):

PR #236, opened 2025-04-22
fix: set topology flag if csi driver advertize it

Madhu-1:
If the admin sets the domain label in the nodeplugin we need to set the topology featuregate to true
in the controller deployment as the feature needs to be enabled on the both provisioner and the nodeplugin , If the domain labels are not set we need to disable the toplogy as its enabled by default.

### DIFF
--- a/internal/controller/driver_controller.go
+++ b/internal/controller/driver_controller.go
@@ -553,6 +553,10 @@ func (r *driverReconcile) reconcileControllerPluginDeployment() error {
 			utils.LeaderElectionRetryPeriodContainerArg(leaderElectionSpec.RetryPeriod),
 		}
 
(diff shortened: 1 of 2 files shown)


nb-ohad (approved):

internal/controller/driver_controller.go
```diff
@@ -553,6 +553,10 @@ func (r *driverReconcile) reconcileControllerPluginDeployment() error {
 			utils.LeaderElectionRetryPeriodContainerArg(leaderElectionSpec.RetryPeriod),
 		}
 
+		// TODO: Move the Topology field from NodePlugin to Driver.Spec
```
Madhu-1, line 556: https://github.com/ceph/ceph-csi-operator/issues/239 to fix the TODO

PR #266, opened 2025-06-26
helm: set appVersion in the helm charts

Madhu-1:
For now we are only setting the version in the helm chart to install the specific version of it, There is another field in the
helm chart that represents the app version. Setting the appVersion similar to version to avoid confusion.

### DIFF
--- a/.github/workflows/publish-helm-charts.yaml
+++ b/.github/workflows/publish-helm-charts.yaml
@@ -43,8 +43,8 @@ jobs:
           echo "Packaging Helm charts..."
           mkdir -p .csi-op-release-packages
 
-          sed -i "s/^version:.*/version: ${VERSION}/" "deploy/charts/ceph-csi-operator/Chart.yaml"
-          sed -i "s/^version:.*/version: ${VERSION}/" "deploy/charts/ceph-csi-drivers/Chart.yaml"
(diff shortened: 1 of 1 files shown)


.github/workflows/publish-helm-charts.yaml
```diff
@@ -43,8 +43,8 @@ jobs:
           echo "Packaging Helm charts..."
           mkdir -p .csi-op-release-packages
 
-          sed -i "s/^version:.*/version: ${VERSION}/" "deploy/charts/ceph-csi-operator/Chart.yaml"
-          sed -i "s/^version:.*/version: ${VERSION}/" "deploy/charts/ceph-csi-drivers/Chart.yaml"
+          sed -i "s/^version:.*/version: ${VERSION}/; s/^appVersion:.*/appVersion: v${VERSION}/" "deploy/charts/ceph-csi-operator/Chart.yaml"
```
leelavg, line 46: Do we want to follow only helm config bumps version and operator changes (like an image) bumps apiVersion or we want less maintenance and have both these same?

Madhu-1: We don't release helm charts separately,  it done along with the operator release,  it's same for both always.

nb-ohad (approved):

PR #269, opened 2025-07-01
driver: deploy csi-addons sidecar in its own DaemonSet

black-dragon74:
This patch introduces a new DaemonSet for csi-addons sidecar that uses pod network instead of the host networking.

### DIFF
--- a/.github/actions/deploy-charts/action.yaml
+++ b/.github/actions/deploy-charts/action.yaml
@@ -83,6 +83,7 @@ runs:
       shell: bash
       run: |
         DRIVER_NAME="${{ inputs.driverType }}.csi.ceph.com" yq eval '.drivers |= map(.enabled = (.name == strenv(DRIVER_NAME)))' -i ./deploy/charts/ceph-csi-drivers/values.yaml
+        yq eval '.drivers |= map(.deployCsiAddons = (.name == "rbd.csi.ceph.com"))' -i ./deploy/charts/ceph-csi-drivers/values.yaml
 
     - name: Install csi driver helm chart
       shell: bash
@@ -94,9 +95,15 @@ runs:
 
     - name: Check csi pods are running
       shell: bash
+      env:
+        DRIVER_TYPE: ${{ inputs.driverType }}
       run: |
+        podCount=2
+        if [ "$DRIVER_TYPE" = "rbd" ]; then
+          podCount=3
+        fi
         for i in {1..180}; do
-          if [ $(kubectl get pods -n csi-driver --field-selector=status.phase=Running --no-headers | wc -l) -eq 2 ]; then
+          if [ $(kubectl get pods -n csi-driver --field-selector=status.phase=Running --no-headers | wc -l) -eq $podCount ]; then
             echo "All CSI pods are running"
             kubectl get pods,deployment,daemonset,replicaset -n csi-driver
             exit 0


internal/controller/driver_controller.go
```diff
@@ -229,16 +229,24 @@ func (r *driverReconcile) reconcile() error {
...
 		r.reconcileLogRotateConfigMap,
 		r.reconcileK8sCsiDriver,
 		r.reconcileControllerPluginDeployment,
 		r.reconcileNodePluginDeamonSet,
 		r.reconcileLivenessService,
-	)
+	}
+
+	// Deploy csi-addons sidecar in its own daemon set which doesn't use host network
+	// This is done only for the RBD driver as of now
```
Madhu-1, line 242: This comment can be removed as this is applicable only to RBD.

internal/controller/driver_controller.go
```diff
@@ -971,6 +979,169 @@ func (r *driverReconcile) controllerPluginCsiAddonsContainerPort() corev1.Contai
...
+			return err
+		}
+		return nil
+	}
+
+	log.Info("Reconciling csi addons nodeplugin deployment")
+
+	opResult, err := ctrlutil.CreateOrUpdate(r.ctx, r.Client, daemonSet, func() error {
+		if err := ctrlutil.SetControllerReference(&r.driver, daemonSet, r.Scheme); err != nil {
+			log.Error(err, "Failed to set owner reference on csi addons nodeplugin deployment")
```
Madhu-1, line 1001: nodeplugin daemonset

internal/controller/driver_controller.go
```diff
@@ -971,6 +979,169 @@ func (r *driverReconcile) controllerPluginCsiAddonsContainerPort() corev1.Contai
...
+
+	if !ptr.Deref(r.driver.Spec.DeployCsiAddons, false) {
+		if err := r.Delete(r.ctx, daemonSet); client.IgnoreNotFound(err) != nil {
+			log.Error(err, "failed to delete csi addons daemonset")
+			return err
+		}
+		return nil
+	}
+
+	log.Info("Reconciling csi addons nodeplugin deployment")
```
Madhu-1, line 997: nodeplugin daemonset.

internal/controller/driver_controller.go
```diff
@@ -971,6 +979,169 @@ func (r *driverReconcile) controllerPluginCsiAddonsContainerPort() corev1.Contai
...
+					DNSPolicy:   corev1.DNSClusterFirstWithHostNet,
+					Tolerations: pluginSpec.Tolerations,
+					Containers: utils.Call(func() []corev1.Container {
+						containers := []corev1.Container{
+							{
+								Name:            "csi-addons",
+								Image:           r.images["addons"],
+								ImagePullPolicy: imagePullPolicy,
+								// We need this in order for this container to be able to access
+								// the sockets created by the privileged provisioner container
```
Madhu-1, line 1049: privileged nodeplugin container

internal/controller/driver_controller.go
```diff
@@ -1157,92 +1328,6 @@ func (r *driverReconcile) reconcileNodePluginDeamonSet() error {
...
-										mounts = append(mounts, utils.LogsDirVolumeMount)
-									}
-									return mounts
-								}),
-								Resources: ptr.Deref(
-									pluginSpec.Resources.Addons,
-									corev1.ResourceRequirements{},
-								),
-							})
-							// Liveness Sidecar Container
```
Madhu-1, line 1211: why liveness is removed?

Madhu-1 (changes requested):
@black-dragon74 can you adjust the operator testing to enable csi-addons and verify extra daemonset for rbd?

internal/controller/driver_controller.go
black-dragon74: This was due to an earlier review at: https://github.com/ceph/ceph-csi-operator/pull/253#discussion_r2156584479

black-dragon74: Ah my bad, fixed :)

black-dragon74:
In my testing, sometimes a deployment deadlock is reached due to strict pod anti-affinity rules (pods are unable to be scheduled on the same node as they are already running).

This is leading to new replicaset always being in pending state, unless we go ahead and delete the old pods manually.

To remedy this we should set maxSurge to 0. Discussing it here before including in this PR.

Madhu-1:
> > This is not related to this PR right?
> 
> Correct. But we could have the updated strategy as a part of this PR itself. Since it is already in the review process.

This is a different problem, it should have its own issue and one PR not as part of this one.

black-dragon74:
> This is a different problem, it should have its own issue and one PR not as part of this one.

Tracked at #273

PR #282, opened 2025-07-24
Deploy CSI-Addons with volume condition reporter

nixpanic:

When a Driver gets annotated with

    addons.csi.ceph.io/volume-condition: true

the CSI-Addons DaemonSet for that driver will get deployed with the extra `--enable-volume-condition=true` commandline flag. This causes the CSI-Addons sidecar to periodically check attached volumes for this Driver and report the condition in the logs of the sidecar and as an Event for the PersistentVolumeClaim.

The volume condition reporter has been merged in csi-addons/kubernetes-csi-addons#831, but a release is still pending. That means the `--enable-volume-condition` commandline flag for the sidecar does not work without replacing the container-image with `quay.io/csiaddons/k8s-sidecar:latest`.

## Future concerns ##

- currently CSI-Addons for NFS is forcefully disabled, this should become optional too

**Checklist:**

  guidelines in the [developer
* [x] [Pending release

### DIFF
--- a/internal/controller/driver_controller.go
+++ b/internal/controller/driver_controller.go
@@ -26,6 +26,7 @@ import (
 	"reflect"
 	"regexp"
 	"slices"
+	"strconv"
 	"strings"
 
 	"github.com/go-logr/logr"
@@ -72,7 +73,10 @@ const (
 const (
 	// Annotation name for ownerref information
 	ownerRefAnnotationKey = "csi.ceph.io/ownerref"
-	logRotateCmd          = `while true; do logrotate --verbose /logrotate-config/csi; sleep 15m; done`
+	// Annotation to enable CSI-Addons volume condition reporter
+	driverCSIAddonsFeatureVolumeCondition = "addons.csi.ceph.io/volume-condition"
+
+	logRotateCmd = `while true; do logrotate --verbose /logrotate-config/csi; sleep 15m; done`
 )
 
 // A regexp used to parse driver's prefix and type from the full name
@@ -236,10 +240,7 @@ func (r *driverReconcile) reconcile() error {
 		r.reconcileControllerPluginDeployment,
 		r.reconcileNodePluginDeamonSet,
 		r.reconcileLivenessService,
-	}
-
-	if r.isRbdDriver() {
-		reconcilers = append(reconcilers, r.reconcileNodePluginDeamonSetForCsiAddons)
+		r.reconcileNodePluginDeamonSetForCsiAddons,
 	}
 
 	// Concurrently reconcile different aspects of the clusters actual state to meet
@@ -981,7 +982,38 @@ func (r *driverReconcile) reconcileNodePluginDeamonSetForCsiAddons() error {
 
 	log := r.log.WithValues("csiAddonsDaemonSetName", daemonSet.Name)
 
-	if !ptr.Deref(r.driver.Spec.DeployCsiAddons, false) {
+	withCsiAddonsDaemonSet := false
+	withCsiAddonsVolumeCondition := false
+
+	if r.isRbdDriver() {
+		withCsiAddonsDaemonSet = ptr.Deref(r.driver.Spec.DeployCsiAddons, false)
+	}
+
+	// check if the driver wants CSI-Addons features
+	if feature := r.driver.GetAnnotations()[driverCSIAddonsFeatureVolumeCondition]; feature != "" {
+		enabled, err := strconv.ParseBool(feature)
+		if err != nil {
+			r.log.Error(
+				err,
+				"Unable to parse annotation on driver.csi.ceph.io",
+				"name",
+				client.ObjectKeyFromObject(&r.driver),
+				driverCSIAddonsFeatureVolumeCondition,
+				feature,
+			)
+			return err
+		}
+
+		withCsiAddonsVolumeCondition = enabled
+		// if the feature is enabled, enable the daemonset too
+		withCsiAddonsDaemonSet = withCsiAddonsDaemonSet || withCsiAddonsVolumeCondition
+	}
+
+	if r.isNfsDriver() {
+		withCsiAddonsDaemonSet = false
+	}
+
+	if !withCsiAddonsDaemonSet {
 		if err := r.Delete(r.ctx, daemonSet); client.IgnoreNotFound(err) != nil {
 			log.Error(err, "failed to delete csi addons daemonset")
 			return err
@@ -1062,6 +1094,7 @@ func (r *driverReconcile) reconcileNodePluginDeamonSetForCsiAddons() error {
 										utils.If(logRotationEnabled, utils.LogToStdErrContainerArg, ""),
 										utils.If(logRotationEnabled, utils.AlsoLogToStdErrContainerArg, ""),
 										utils.If(logRotationEnabled, utils.LogFileContainerArg("csi-addons"), ""),
+										utils.If(withCsiAddonsVolumeCondition, utils.CsiAddonsVolumeConditionArg, ""),
 									},
 								),
 								Ports: []corev1.ContainerPort{
@@ -1079,6 +1112,12 @@ func (r *driverReconcile) reconcileNodePluginDeamonSetForCsiAddons() error {
 									}
 									if logRotationEnabled {
 										mounts = append(mounts, utils.LogsDirVolumeMount)
+									}
+									if withCsiAddonsVolumeCondition {
+										mounts = append(mounts,
+											utils.PluginMountDirVolumeMount(kubeletDirPath),
+											utils.PodsMountDirVolumeMount(kubeletDirPath),
+										)
 									}
 									return mounts
 								}),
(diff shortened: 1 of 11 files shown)


internal/controller/driver_controller.go
```diff
@@ -238,7 +245,7 @@ func (r *driverReconcile) reconcile() error {
 		r.reconcileLivenessService,
 	}
 
-	if r.isRdbDriver() {
+	if r.wantsCsiAddonsDaemonSet() {
```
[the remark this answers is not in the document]
Madhu-1: With this new change i think we should skip this and add it to reconcilers by default and take care of it inside `reconcileNodePluginDeamonSetForCsiAddons` whether to create or delete.

internal/controller/driver_controller.go
```diff
@@ -72,7 +73,10 @@ const (
 const (
 	// Annotation name for ownerref information
 	ownerRefAnnotationKey = "csi.ceph.io/ownerref"
-	logRotateCmd          = `while true; do logrotate --verbose /logrotate-config/csi; sleep 15m; done`
+	// Annotation to enable CSI-Addons volume condition reporter
+	driverCSIAddonsFeatureVolumeCondition = "feature.csi.ceph.io/volume-condition"
```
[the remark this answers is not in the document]
nixpanic: That can be done, but a single annotation per feature might be easier for users?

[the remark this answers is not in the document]
nixpanic: From Madhu over Slack:

> csi.ceph.io/feature/volume-condition should good for better grouping of features

I'll update this PR and use that.

internal/controller/driver_controller.go
```diff
@@ -313,6 +317,24 @@ func (r *driverReconcile) LoadAndValidateDesiredState() error {
...
+				"name",
+				client.ObjectKeyFromObject(&r.driver),
+				driverCSIAddonsFeatureVolumeCondition,
+				feature,
+			)
+			return err
+		}
+
+		r.withCsiAddonsVolumeCondition = enabled
+	}
```
Madhu-1, line 336: This check can be moved to inside reconcileNodePluginDeamonSetForCsiAddons we dont need a new variable as well `withCsiAddonsVolumeCondition`, please add a check for this only if the driver is cephFS

internal/controller/driver_controller.go
```diff
@@ -1508,6 +1548,19 @@ func (r *driverReconcile) isNfsDriver() bool {
 	return r.driverType == NfsDriverType
 }
 
+func (r *driverReconcile) wantsCsiAddonsDaemonSet() bool {
```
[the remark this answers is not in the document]
Madhu-1: We can remove this helper as its used in only one place and directly have the logic in reconcileNodePluginDeamonSetForCsiAddons?

internal/controller/driver_controller.go
```diff
@@ -981,7 +1003,7 @@ func (r *driverReconcile) reconcileNodePluginDeamonSetForCsiAddons() error {
 
 	log := r.log.WithValues("csiAddonsDaemonSetName", daemonSet.Name)
 
-	if !ptr.Deref(r.driver.Spec.DeployCsiAddons, false) {
+	if !r.wantsCsiAddonsDaemonSet() {
```
Madhu-1, line 1006:
```suggestion
wantCSIAddons := r.IsRBD() && ptr.Deref(r.driver.Spec.DeployCsiAddons, false)

if r.IsCephFS() {
	if feature := r.driver.GetAnnotations()[driverCSIAddonsFeatureVolumeCondition]; feature != "" {
		var err error
		wantCSIAddons, err = strconv.ParseBool(feature)
		if err != nil {
			r.log.Error(
				err,
				"Unable to parse annotation on driver.csi.ceph.io",
				"name", client.ObjectKeyFromObject(&r.driver),
				driverCSIAddonsFeatureVolumeCondition,
				feature,
			)
			return err
		}
	}
} else if r.IsNFS() {
	wantCSIAddons = false
}
```

internal/controller/driver_controller.go
nixpanic: Well, even for RBD the feature can be enabled. By default RBD will have the standard CSI-Addons DaemonSet, but with the annotation also the volume condition of RBD volumes will be reported.

I can move it into `reconcileNodePluginDeamonSetForCsiAddons`, and check for `!r.isNfsDriver()` to prevent issues with it there.

Does that make sense to you too, @Madhu-1 ?

Madhu-1: @nixpanic it wont work for RBD right, because its missing RBAC's isnt it? if its enabled and works that should be fine.

internal/controller/driver_controller.go
```diff
@@ -981,7 +982,36 @@ func (r *driverReconcile) reconcileNodePluginDeamonSetForCsiAddons() error {
 
 	log := r.log.WithValues("csiAddonsDaemonSetName", daemonSet.Name)
 
-	if !ptr.Deref(r.driver.Spec.DeployCsiAddons, false) {
+	// FIXME: Spec.DeployCsiAddons defaults to true, enabling CSI-Addons for CephFS by default
+	withCsiAddonsDaemonSet := r.isRbdDriver() // || ptr.Deref(r.driver.Spec.DeployCsiAddons, false)
```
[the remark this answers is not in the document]
nixpanic: @Madhu-1 , added this now too. Thanks!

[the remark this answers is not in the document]
nixpanic: tested it, and the DaemonSet gets deleted when setting `spec.deployCsiDriver=false`:
```
$ oc -n openshift-storage get driver/openshift-storage.rbd.csi.ceph.com -oyaml
apiVersion: csi.ceph.io/v1
kind: Driver
metadata:
  creationTimestamp: "2025-07-25T09:19:00Z"
  generation: 2
  name: openshift-storage.rbd.csi.ceph.com
  namespace: openshift-storage
  ...
spec:
  deployCsiAddons: false
```

internal/controller/driver_controller.go
```diff
@@ -981,7 +982,41 @@ func (r *driverReconcile) reconcileNodePluginDeamonSetForCsiAddons() error {
 
 	log := r.log.WithValues("csiAddonsDaemonSetName", daemonSet.Name)
 
-	if !ptr.Deref(r.driver.Spec.DeployCsiAddons, false) {
+	// FIXME: Spec.DeployCsiAddons defaults to true, causes CSI-Addons for CephFS enabled by default
+	withCsiAddonsDaemonSet := false // ptr.Deref(r.driver.Spec.DeployCsiAddons, false)
```
Madhu-1, line 986: @nixpanic need to remove the FixME and the comment as no more defaulting to true

internal/controller/driver_controller.go
```diff
@@ -981,7 +982,41 @@ func (r *driverReconcile) reconcileNodePluginDeamonSetForCsiAddons() error {
...
+	withCsiAddonsVolumeCondition := false
+
+	// FIXME: this workaround allows a user to disable the DaemonSet for RBD
```
Madhu-1, line 989: @nixpanic This is not a workaround i think its the actual behaviour

internal/controller/driver_controller.go
```diff
@@ -981,7 +982,41 @@ func (r *driverReconcile) reconcileNodePluginDeamonSetForCsiAddons() error {
...
+			)
+			return err
+		}
+
+		withCsiAddonsVolumeCondition = enabled
+		// if the feature is enabled, enable the daemonset too
+		withCsiAddonsDaemonSet = withCsiAddonsDaemonSet || withCsiAddonsVolumeCondition
+	}
+
+	// FIXME: explicitly disable for NFS, depends on #284
```
Madhu-1, line 1014: Remove FiXME as we dont want to deploy this for NFS thats why we have this code?

Madhu-1 (commented):
small nits, LGTM

internal/controller/driver_controller.go
nixpanic: Well, in the future we surely want to be able to enable this for NFS too?

internal/controller/driver_controller.go
nixpanic:
```go
ptr.Deref(r.driver.Spec.DeployCsiAddons, false)
```
returns `true` even when `.Spec.DeployCsiAddons` is not set in the CR? I think that is still the case, or at least I see that in my test environment.

internal/controller/driver_controller.go
nixpanic: Yeah, well, using `ptr.Deref(r.driver.Spec.DeployCsiAddons, false)` should be done above:

```go
withCsiAddonsDaemonSet := ptr.Deref(r.driver.Spec.DeployCsiAddons, false)
```

and then this if statement can be removed. But that requires `ptr.Deref(r.driver.Spec.DeployCsiAddons, false)` to return `false` if it isn't set.

Madhu-1 (approved):

internal/controller/driver_controller.go
Madhu-1: We will rework on this with proper API change once its becomes to Beta state.

PR #290, opened 2025-08-05
doc: use csi.ceph.io/v1 instead of v1alpha1

nixpanic:
The API moved to v1, but the quickstart and other documents still
referred to v1alpha1.


Madhu-1 (commented):
Changes LGTM, can you please make similar change in kube-installation file as well?

Madhu-1 (approved):

PR #291, opened 2025-08-07
csi: handle setmetatadata false condition

Nikhil-Ladha:

ceph-csi recently made the default value `setmetadata` flag as true, which means if we don't provide the flag it will still be set as true.

Therefore, we should explicitly set the flag to false if the argument is overriden through the driver CR.

### DIFF
--- a/internal/utils/csi.go
+++ b/internal/utils/csi.go
@@ -425,7 +425,7 @@ func TypeContainerArg(t string) string {
 	}
 }
 func SetMetadataContainerArg(on bool) string {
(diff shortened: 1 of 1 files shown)


leelavg:
I don't think all flags should be treated equal or am I missing your statement that fencing issue is caused by this flag?

we made this flag always to true based on https://github.com/red-hat-storage/ocs-client-operator/pull/378#issuecomment-2911509191 and having this flag might also helps in getting metrics in hub side even when csi is running in client.

Issue #305, opened 2025-08-22
Add option to set imagePullSecrets


FlorisFeddema:
# Describe the feature you'd like to have #

Currently there is no option to add an imagePullSecret to the driver plugins.
Also the helm chart does not provide a way to add this for the 

## Who is the end user and what is the use case where this feature will be valuable? ##
  
We cache all of the images in our own registry and override the image repository with our own. Since this container registry has authentication we would like to provide the credentials via the imagePullSecret field.

# How will we know we have a good solution? (acceptance criteria) #

Once it is possible to set an imagePullSecret for all pods in the application and thus all images can be pulled from a private registry.

# Additional context #

PR #314, opened 2025-09-03
support custom imagePullSecrets in all ServiceAccounts

leelavg:
sample usage:
IMAGE_PULL_SECRETS="secret-1" make build-installer
helm template csi deploy/charts/ceph-csi-drivers --set 'imagePullSecrets={secret-1}'
helm template csi deploy/charts/ceph-csi-operator --set 'imagePullSecrets={secret-1}'

fixes: #305

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -9,6 +9,8 @@ NAME_PREFIX ?= ceph-csi-operator-
 NAMESPACE ?= $(NAME_PREFIX)system
 # A comma separated list of namespaces for operator to cache objects from
 WATCH_NAMESPACE ?= ""
+# A comma separated list of imagepullsecret names that will be added to all CSI deployments and daemonsets
+IMAGE_PULL_SECRETS ?=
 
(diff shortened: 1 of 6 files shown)


nb-ohad:
@leelavg I am not sure how does a change to the make file solves the issue. If I understand correctly it will require to build a custom version of the operator. Or at least run some make commends locally.

leelavg:
updated.

I realized we have two sets of helms charts, csi-op as a superset of csi-drivers w/ duplication, need to know the reason from @Madhu-1 maybe we could just club them using subcharts depedency 🤔

leelavg:
I'm awaiting to get #321 merged to take in the split, csi-rbac is already covered and for csi-op I'll invoke helmify with `-image-pull-secrets` which adds the field directly to the deployment and creates a configurable field in `values.yaml`.

Madhu-1:
Closing this one as its fixed now.

PR #321, opened 2025-09-19
helm: remove csi RBAC from the operator

Madhu-1:
CSI template and RBAC's are shipped as another charts. we dont need to add the csi RBAC to the operator charts.


Madhu-1:
> What chart is the rbac in? Does Rook need to add a helm dependency on it or does the csi operator chart have a dependency on it?

@travisn the CSI RBAC was part of both charts. am removing it from the operator chart so that it will be present only in the csi-driver charts

travisn:
> > What chart is the rbac in? Does Rook need to add a helm dependency on it or does the csi operator chart have a dependency on it?
> 
> @travisn the CSI RBAC was part of both charts. am removing it from the operator chart so that it will be present only in the csi-driver charts

Rook doesn't currently install the csi-driver chart, only the csi-operator chart. Can this be factored out as a helm library so the rbac can still come with the csi-operator chart?

travisn:
To summarize our discussion:
- In the next minor release of Rook (v1.19), Rook will add a helm dependency on the csi drivers chart. 
- This RBAC change will not be made until the CSI operator 1.0 release (which is when Rook v1.19 will consume it)
- Note that csi drivers chart is for creating the csi CRs (and is not related to the old csi drivers chart)

Madhu-1:
@travisn @subhamkrai are we good to merge this one?

subhamkrai:
> @travisn @subhamkrai are we good to merge this one?

Yeah, looks good to me with all the Rook testing

travisn (commented):
Sounds good to merge, and we will follow up with Rook e2e testing as well

travisn (approved):

Madhu-1:
I will rework on the whole RBAC where the ClusterRoles will remain with the operator and we will only create the Rolebinding in the driver chart later

Issue #312, opened 2025-09-02
Do not add pod antiaffinity to controller plugin deployment if replica 1


travisn:
# Describe the bug #

Pod anti-affinity is currently always required for the controller plugin as implemented [here](https://github.com/ceph/ceph-csi-operator/blob/2587c60f521cc5457b7e76a9b1a3806190ce3c86/internal/controller/driver_controller.go#L1639-L1644). In single-node clusters, this can cause upgrade issues since the new pod does not have another node to start on before the current pod is deleted. 

If only one replica is configured, the suggestion is to not add any pod antiaffinity so that upgrades will not be blocked. For example, when replicas 1 is requested in the operatorConfig CR (or specific drivers):
```yaml
  spec:
    driverSpecDefaults:
      controllerPlugin:
        replicas: 1
```
See the rook issue where this is described in more detail. 
https://github.com/rook/rook/issues/16419

leelavg:
> If replicas==1, then there is no need for pod anti-affinity since there is only one pod.
- if there is only a single node then no need of affinity so I thought affinity would still make sense even if replica is 1 but num of nodes is >1 ensuring the upgrade pod also doesn't end up on running node

ack for the rest!

black-dragon74:
@leelavg I guess this is one more case where an alternate `updateStrategy` makes sense?

leelavg:
> @leelavg I guess this is one more case where an alternate `updateStrategy` makes sense?
- I don't think so. They serve different purposes. You can set the deployment strategy to delete the older one during an upgrade but it removes the redundancy.

black-dragon74:
I can send a PR. We discussed to change the strategy to Recreate but then that comes at the cost of redundancy. 

We can have maxSurge: 0 and maxUnavailable: 1 too (better IMO).

#273 is open for that but we need to agree on one approach first.

Madhu-1:
lets change the strategy to Recreate as its long proven with cephcsi as default driver and doesnt have any issues

Madhu-1:
> [@Madhu-1](https://github.com/Madhu-1) - shouldn't the default apply to the helm chart too? Currently, both Driver and OperatorConfig set the controllerPlugin.deploymentStrategy = {}, so they get k8s defaults rather than the operator default.
> 

if this is happening, we should fix this one. @black-dragon74 Can you please check this one.

[...]

Issue #273, opened 2025-07-17
Update the deployment strategy for CSI deployments


black-dragon74:
# Describe the bug #

When the deployments managed by the operator are updated and the pod placement in the same node,
it leads to the new pods always being in the pending state due to the way we have defined the `updateStrategy` and `podAntiAffinity` rules.

The strategy of maxSurge > 0 instructs the scheduler to create new pods before it evicts older ones. The new pods will never be in running state as the anti-affinity rules prevent that.

As a result, we reach a deadlock. _The new pods can't start because of the old pods, and the old pods won't be removed until the new one starts._

To remedy this, we can have `maxSurge: 0` and `maxUnavailable: 1`

black-dragon74:
That would be quite disruptive (all pods will be deleted first) and we would have downtime (albeit a few seconds). In place rollouts make the most sense (delete one, recreate it, repeat for remaining replicas).

Madhu-1:
Recreate is the strategy we are using from many years(almost 5) by default in Rook

Madhu-1:
leaderelection timeout is i think 2.5 minutes, if you loose the leader even in case of the RolingUpdate its the same no leader will be present to serve the request. the Recreate is better strategy as it will bring back the leader earlier than rolingupdate

black-dragon74:
> Recreate is the strategy we are using from many years(almost 5) by default in Rook

_Intrigued by why we chose rolling updates for csi-op. Recreate is obviously much simpler and straightforward_

Madhu-1:
> for `maxSurge: 0` you probably want to wait for [kubernetes/kubernetes#107920](https://github.com/kubernetes/kubernetes/issues/107920)

Thanks @sebhoss for the link, this is long way. we should look for alternative like Recreate which is opt for ceph-csi-operator

black-dragon74:
Fixed with: #322

PR #322, opened 2025-09-23
driver: Change controller-plugin deployment strategy to Recreate

black-dragon74:

This patch changes the default update strategy for controller plugin to Recreate from RollingUpdate.

This helps fix the situation where we end up in a scheduling deadlock due to affinity and anti affinity rules.

Fixes: #312, #273


sshambar:
@Madhu-1 Shouldn't the CRDs documentation for Driver be updated with the new default too?

Madhu-1:
> @Madhu-1 Shouldn't the CRDs documentation for Driver be updated with the new default too?

@sshambar yes its required, i think we missing updating it in the types comment. @black-dragon74 PTAL

Issue #339, opened 2025-10-05
ceph-csi-drivers chart does not properly quote monitor addresses


lilioid:
# Describe the bug #

When using the *ceph-csi-drivers* chart with monitor addresses that include brackets (e.g. IPv6 addresses or the address format exposed by `ceph mon dump`), the chart breaks because it generates invalid YAML. This is because the [cephConnection.yaml template](https://github.com/ceph/ceph-csi-operator/blob/ca37138d4e9001525abf53a593a7aea5e18e0aaa/deploy/charts/ceph-csi-drivers/templates/cephConnection.yaml#L12) uses `{{ . }}` instead of `{{ . | quote }}`.

# Environment details #

- Image/version of ceph-csi-operator : `0.4.1`
- Image/version of Ceph-CSI driver : `0.4.1`
- Kubernetes cluster version : irrelevant
- Ceph cluster version : irrelevant

# Steps to reproduce #

1. Call `helm template` with a values file that includes `cephConnections.*.monitors` with `[v2:[abc::123]:3300/0,v1:[abc:456]:6789/0]` (i.e. the format exposed by `ceph mon dump`)

# Actual results #

Helm fails with the following error:

```
Error: YAML parse error on ceph-csi-drivers/templates/cephConnection.yaml: error converting YAML to JSON: yaml: line 7: did not find expected ',' or ']'
```

# Expected behavior #

Helm should be able to generate valid kubernetes manifests

# Logs #

# Additional context #

PR #345, opened 2025-10-06
helm: quote ceph mon addresses in CephConnection

black-dragon74:

This patch fixes an issue where an invalid YAML was rendered in cases of IPv6 addresses.

The error was due to missing quotes around the addresses.

Fixes: #339

### DIFF
--- a/deploy/charts/ceph-csi-drivers/templates/cephConnection.yaml
+++ b/deploy/charts/ceph-csi-drivers/templates/cephConnection.yaml
@@ -9,7 +9,7 @@ metadata:
 spec:
   monitors:
   {{- range $cephConnection.monitors }}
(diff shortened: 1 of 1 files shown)


Madhu-1:
@black-dragon74 can we update https://github.com/ceph/ceph-csi-operator/blob/0b2704e60057e66da53535d3c147436245e6225e/.github/workflows/test-charts.yaml#L38-L46 this to set this in the helm installation as well. so that this will also gets tested by CI.

Madhu-1:
https://github.com/ceph/ceph-csi-operator/blob/0b2704e60057e66da53535d3c147436245e6225e/.github/actions/deploy-charts/action.yaml#L82-L91 need to be added here

PR #368, opened 2025-12-12
cephfs: deploy csi-addons daemonset

iPraveenParihar:

cephfs driver now supports GetFenceClient RPC call and
need csi-addons daemonset to be deployed for cephfs driver.

Depends-on: https://github.com/ceph/ceph-csi/pull/5828


Madhu-1:
Remove the volume condition check and use csiaddon flag for it. Let's not have 2 different ways of doing it.

Issue #384, opened 2026-01-16
csi-controller-manager helm chart does not manage namespace


urbaman:
# Describe the bug #

The csi-controller-manager via helm gets deployed in the default namespace instead of the rightful rook-ceph ns: https://github.com/rook/rook/discussions/16932#discussion-9353735

# Steps to reproduce #

helm template --create-namespace --namespace rook-ceph rook-ceph rook-release/rook-ceph > rookCephManifest.yaml

# Actual results #

The csi-controller-manager (and all of the other resources, like serviceAccounts) got created in the default namespace

# Expected behavior #

The csi-controller-manager (and all of the other resources, like serviceAccounts) gets created in the rook-ceph namespace

Madhu-1:
fixed in https://github.com/ceph/ceph-csi-operator/pull/392

PR #391, opened 2026-01-28
Explicitly set namespace for Ceph CSI operator/driver resources

degorenko:
Add explicitly namespace name for CephCSI operator resources to have an ability specify it during helm install.

Fixes: #384


degorenko:
I guess issue that kustomize build with `ceph-csi-operator-system` namespace, which is preserved by helmify itself (so `-preserve-ns` flag is actually working fine)

Madhu-1:
@degorenko i have sent PR #392 to fix it for the operator for now, will take care of the helm charts later when Rook adopts to it.

Madhu-1:
fixed in https://github.com/ceph/ceph-csi-operator/pull/392

PR #449, opened 2026-04-07
feat: add extraArguments support for CSI drivers

varet80:
I am submitting an initial PR which will enable extraArgs on CSI Drivers.

In order to support configureable arguments. In our case, we need to be able to modify the following parameters on csi driver:

```
 --rbdhardmaxclonedepth=<int>
 --rbdsoftmaxclonedepth=<int>
```

* [X] **Commit Message Formatting**: Commit titles and messages follow
  guidelines in the [developer
* [X] Reviewed the developer guide on [Submitting a Pull
* [X] [Pending release
* [X] Documentation has been updated, if necessary.
* [X] Unit tests have been added, if necessary.
* [X] Integration tests have been added, if necessary.


Madhu-1:
@varet80 unfortunately we cannot add all the configuration to the API spec,we need to come up with a design on how to avoid deprecation or removal of new fields when they get removed in cephcsi

varet80:
This is a Critical setting on CSI in order to be able to configure flattening depth levels.
It is not a value which will be deprecated but also not everybody configures this.

The alternative would be, the ability to add any CSI driver command arguments through the operator. Would that be an option as alternative?

Madhu-1:
> This is a Critical setting on CSI in order to be able to configure flattening depth levels. It is not a value which will be deprecated but also not everybody configures this.
> 
> The alternative would be, the ability to add any CSI driver command arguments through the operator. Would that be an option as alternative?

@varet80 yes i am thinking the same to keep it opaque we dont need to add any new API when there is a new CLI agrument in cephcsi

varet80:
@Madhu-1 I have modified my pr to support the extraArg method.
Can you kindly look the new suggested method?

Madhu-1:
> @Madhu-1 I have modified my pr to support the extraArg method. Can you kindly look the new suggested method?

@varet80 i was thinking in similar line but with little modification. we need to have a map where the key should be the container name and the values are list of keys which we need to apply to the container, with this we will also be able to cover different usecase where someone want to enable or disable certain alpha features

varet80:
> > @Madhu-1 I have modified my pr to support the extraArg method. Can you kindly look the new suggested method?
> 
> @varet80 i was thinking in similar line but with little modification. we need to have a map where the key should be the container name and the values are list of keys which we need to apply to the container, with this we will also be able to cover different usecase where someone want to enable or disable certain alpha features

Can you share an example on the csi driver consumption (the snippet of the spec) to understand the structure?
that way I could modify my pr based on that example.

Madhu-1:
I am thinking something like below as its advance configurations, we can think about overriding what default set by csi-operator later. 

```yaml
apiVersion: csi.ceph.io/v1
kind: Driver
metadata:
  labels:
    app.kubernetes.io/name: ceph-csi-operator
    app.kubernetes.io/managed-by: kustomize
  name: rbd.csi.ceph.com
spec:
    ...
    # Extra arguments for controller plugin containers
    # The key is the container name, the value is a list of CLI arguments
    # The operator will not override the default CLI argument values its setting
    # This is more for customizing what comes as default with cephcsi 
    extraArgs:
      # Configure clone flattening parameters for the main csi-rbdplugin container
      csi-rbdplugin:
        - "--rbdhardmaxclonedepth=8"
        - "--rbdsoftmaxclonedepth=4"
      # Enable additional features for the csi-provisioner sidecar
      csi-provisioner:
        - "--extra-create-metadata=true"
        - "--prevent-volume-mode-conversion=true"

      # Configure timeout for csi-attacher sidecar
      csi-attacher:
        - "--timeout=150s"  # Override default timeout

      # Configure csi-resizer with custom settings
      csi-resizer:
        - "--handle-volume-inuse-error=false"
       
      # Configure csi-snapshotter with additional metadata
      csi-snapshotter:
        - "--extra-create-metadata=true"
```

varet80:
> I am thinking something like below as its advance configurations, we can think about overriding what default set by csi-operator later.
> 

I have made the modifications the way you suggested.
and we are testing it on our environment.

internal/utils/csi.go
```diff
@@ -486,3 +487,37 @@ func DomainLabelsContainerArg(options []string) string {
 func TopologyContainerArg(topology bool) string {
 	return fmt.Sprintf("--feature-gates=Topology=%t", topology)
 }
+
+func ExtraDriverArgsContainerArgs(args map[string]string) []string {
```
Madhu-1: This is not used anywhere, can we remove it?

Madhu-1 (commented):
@varet80 did you get a chance to test the changes and working as expected?

internal/utils/csi.go
varet80: I have resolved this.

varet80:
> @varet80 did you get a chance to test the changes and working as expected?

Yes we did test both the early version and the latest. Works as expected for us.

image

api/v1/driver_types.go
```diff
@@ -376,6 +376,15 @@ type DriverSpec struct {
 	// Set mount options to use when using the Fuse client
 	//+kubebuilder:validation:Optional
 	FuseMountOptions map[string]string `json:"fuseMountOptions,omitempty"`
+
+	// Extra arguments for containers in the driver deployment.
```
Madhu-1: @varet80 can we move this to ControllerPluginSpec and NodePluginSpec, the reason is that we will have same container name for both deployment and deaemonset, we should be able to provide different arguments for both some times.

api/v1alpha1/driver_types.go
```diff
@@ -371,6 +371,15 @@ type DriverSpec struct {
...
 	FuseMountOptions map[string]string `json:"fuseMountOptions,omitempty"`
+
+	// Extra arguments for containers in the driver deployment.
+	// The key is the container name, the value is a list of CLI arguments.
+	// The operator will not override the default CLI argument values it's setting.
+	// This is more for customizing what comes as default with cephcsi.
+	// Examples of container names: csi-rbdplugin, csi-provisioner, csi-attacher,
+	// csi-resizer, csi-snapshotter, csi-registrar, liveness-prometheus, etc.
+	//+kubebuilder:validation:Optional
+	ExtraArgs map[string][]string `json:"extraArgs,omitempty"`
```
Madhu-1: lets not add it to alphav1, its already deprecated and removing the support for the same

Madhu-1 (commented):
small nits, LGTM

api/v1/driver_types.go
varet80: Makes sense. I did my best to make this change! can you check it?

api/v1alpha1/driver_types.go
varet80: done!

Madhu-1:
@varet80 can you please fix the DCO and squash all commits into 1?

Madhu-1:
@varet80 CI is still failing, let me know once its fixed, we can get it merged after that

varet80:
@Madhu-1 I had to fix some conflicts too. Is it ready that way for workflows to be run?

deploy/charts/ceph-csi-drivers/values.yaml
```diff
@@ -110,6 +110,8 @@ operatorConfig:
       privileged: false
       # -- List of tolerations for the controller plugin (default: [])
       tolerations: []
+      # -- Extra driver arguments (default: {})
+      extraDriverArgs: {}
```
Madhu-1: In helm chart we are setting only in the controllerplugin not in nodeplugin, can you  please add the same to the nodeplugin spec as well and also update the tempate, please take care of same for operatorConfig as well. somehow i missed this one

Madhu-1 (dismissed):

Madhu-1:
@varet80 i made the changes and push it to this PR. once CI passes will merge this one, Thank you for taking this up!

Madhu-1 (dismissed):

Madhu-1 (approved):

varet80:
@Madhu-1 sorry for the late reply, it was evening for me (European timezone) and thank you for fixing it, I just started looking into it.

Madhu-1:
> @Madhu-1 sorry for the late reply, it was evening for me (European timezone) and thank you for fixing it, I just started looking into it.

no problem, Thanks for taking time to work on it

PR #451, opened 2026-04-09
fix: explicitly set boolean flag values for both true and false

### LINKED ISSUE #380: Operator fails to set --setmetadata=false
# Describe the bug #

The Driver.spec.enableMetadata is false by default, but the reconciler only applies --setmetadata=true if enableMetadata is true -- however --setmetadata is true by default, so the key has no effect.

# Environment details #

- Image/version of ceph-csi-operator : 0.4.1
- Image/version of Ceph-CSI driver : v3.15.0
- Kubernetes cluster version : v1.33.5
- Ceph cluster version : 19.2.3

# Steps to reproduce #

1. Set a Driver.spec.enableMetadata: false (or don't specify it)
2. Apply the driver, and examine the ctrlplugin controller deployment

# Actual results #

The csi-cephfsplugin container args doesn't include '--setmetadata={bool}'

# Expected behavior #

The csi-cephfsplugin container args should include '--setmetadata=false' 

# Additional context #

For ref: `kubectl exec <ctrlplugin-pod> -- /usr/local/bin/cephcsi -h` provides:
```
  -setmetadata
        set metadata on the volume (default true)
```

iPraveenParihar:

```
SetMetadataContainerArg and SetFencingContainerArg returned an empty string when false,
relying on ceph-csi defaults. If ceph-csi changes its defaults, users who explicitly set these to
false in their Driver CR would be silently ignored. Use fmt.Sprintf with %t to always pass the
explicit value.
```

See: https://github.com/ceph/ceph-csi/pull/6218#issuecomment-4211665029

Fixes: https://github.com/ceph/ceph-csi-operator/issues/380

**Checklist:**

### DIFF
--- a/internal/utils/csi.go
+++ b/internal/utils/csi.go
@@ -424,10 +424,10 @@ func TypeContainerArg(t string) string {
 	}
 }
 func SetMetadataContainerArg(on bool) string {
-	return If(on, "--setmetadata=true", "")
+	return fmt.Sprintf("--setmetadata=%t", on)
 }
 func SetFencingContainerArg(on bool) string {
-	return If(on, "--enable-fencing=true", "")
+	return fmt.Sprintf("--enable-fencing=%t", on)
 }
 func TimeoutContainerArg(timeout int) string {
 	return fmt.Sprintf("--timeout=%ds", timeout)


Madhu-1 (approved):

nixpanic (approved):
lgtm, thanks!

PR #454, opened 2026-04-17
deprecate: driver.spec.EnableMetadata field

iPraveenParihar:

```
The EnableMetadata field and --setmetadata flag are no longer used
by the ceph-csi driver. Remove the field from the DriverSpec API,
generated deepcopy, CRD manifests, and Helm chart values/templates.
```

See: https://github.com/ceph/ceph-csi/pull/6225

**Checklist:**

### DIFF
--- a/api/v1/driver_types.go
+++ b/api/v1/driver_types.go
@@ -291,9 +291,7 @@ type DriverSpec struct {
 	//+kubebuilder:validation:Optional
 	ClusterName *string `json:"clusterName,omitempty"`
 
-	// Set to true to enable adding volume metadata on the CephFS subvolumes and RBD images.
-	// Not all users might be interested in getting volume/snapshot details as metadata on CephFS subvolume and RBD images.
-	// Hence enable metadata is false by default.
+	// Deprecated: This field is no longer used by the ceph-csi driver and will be ignored.
 	//+kubebuilder:validation:Optional
 	EnableMetadata *bool `json:"enableMetadata,omitempty"`
 
--- a/deploy/charts/ceph-csi-drivers/values.yaml
+++ b/deploy/charts/ceph-csi-drivers/values.yaml
@@ -54,8 +54,6 @@ operatorConfig:
       name: ""
     # -- Cluster name identifier (default: "")
     clusterName: ""
-    # -- Flag to enable metadata (default: false)
-    enableMetadata: false
     # -- gRPC timeout in seconds (default: 30)
     grpcTimeout: 30
     # -- Snapshot policy (options: none, volumeGroupSnapshot, volumeSnapshot) (default: "none")
@@ -141,8 +139,6 @@ drivers:
       name: ""
     # -- Cluster name identifier (default: "")
     clusterName: ""
-    # -- Flag to enable metadata (default: false)
-    enableMetadata: false
     # -- gRPC timeout in seconds (default: 30)
     grpcTimeout: 30
     # -- Snapshot policy (options: none, volumeGroupSnapshot, volumeSnapshot) (default: "none")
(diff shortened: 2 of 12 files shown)


api/v1/driver_types.go
```diff
@@ -291,12 +291,6 @@ type DriverSpec struct {
 	//+kubebuilder:validation:Optional
 	ClusterName *string `json:"clusterName,omitempty"`
 
-	// Set to true to enable adding volume metadata on the CephFS subvolumes and RBD images.
-	// Not all users might be interested in getting volume/snapshot details as metadata on CephFS subvolume and RBD images.
-	// Hence enable metadata is false by default.
-	//+kubebuilder:validation:Optional
-	EnableMetadata *bool `json:"enableMetadata,omitempty"`
```
travisn, line 298: Removing CRD settings is a breaking change. Consider leaving this setting here, just deprecate it and only remove the internal code that calls the `--setmetadata` flag.

iPraveenParihar: I was thinking it is okay to remove since we are still at pre v1.0.0 release?

Madhu-1: Lets not remove the CRD setting, we can mark this as deprecated and add a notice that the feature is enabled by default in csi now

travisn: > I was thinking it is okay to remove since we are still at pre v1.0.0 release?

Since we are shipping already in stable upstream and downstream releases, we need to anyway treat it as v1.

iPraveenParihar: understood, will add deprecation notice and keep the field.

deploy/charts/ceph-csi-drivers/values.yaml
```diff
@@ -54,7 +54,7 @@ operatorConfig:
       name: ""
     # -- Cluster name identifier (default: "")
     clusterName: ""
-    # -- Flag to enable metadata (default: false)
+    # -- [Deprecated] This field is no longer used and will be ignored (default: false)
```
travisn: Removing a setting from a helm chart is not a breaking change. Any extra values in the values.yaml would be ignored. So you could go ahead and remove the setting from values.yaml

internal/controller/driver_controller.go
```diff
@@ -1704,8 +1702,8 @@ func mergeDriverSpecs(dest, src *csiv1.DriverSpec) {
 	if dest.ClusterName == nil {
 		dest.ClusterName = src.ClusterName
 	}
-	if dest.EnableMetadata == nil {
-		dest.EnableMetadata = src.EnableMetadata
+	if dest.EnableMetadata == nil { //nolint:staticcheck // backward compatibility for deprecated field
```
travisn: I would think this is not necessary. Backward compatibility is to avoid breaking API changes, but we don't need to keep code that preserves the setting.

travisn (changes requested):

travisn (approved):
LGTM

api/v1/driver_types.go
```diff
@@ -291,9 +291,8 @@ type DriverSpec struct {
 	//+kubebuilder:validation:Optional
 	ClusterName *string `json:"clusterName,omitempty"`
 
-	// Set to true to enable adding volume metadata on the CephFS subvolumes and RBD images.
-	// Not all users might be interested in getting volume/snapshot details as metadata on CephFS subvolume and RBD images.
-	// Hence enable metadata is false by default.
+	// Deprecated: This field is no longer used by the ceph-csi driver and will be ignored.
+	// It will be removed in a future release.
```
Madhu-1: can you remove this line, we have not decided to remove it or not for now.

Madhu-1 (approved):

PR #478, opened 2026-05-25
helm: guard deprecated enableMetadata against nil values

### LINKED ISSUE #473: helm driver chart requires enableMetadata to be set
# Describe the bug #
When installing/upgrade the ceph-csi-drivers chart, using a values.yaml that doesn't include the deprecated operatorConfig.driverSpecDefaults.enableMeta, I receive the following error:

# Environment details #

- Image/version of ceph-csi-operator : v1.0.0
- Image/version of Ceph-CSI driver : v3.16.2
- Kubernetes cluster version : 1.36
- Ceph cluster version : 20.2.1

# Steps to reproduce #

1. Install ceph-csi-operator 1.0.0 with a values.yaml without an operatorConfig section, or some minimal config such as:
```
operatorConfig:
  driverSpecDefaults:
    log:
      verbosity: 1
```

# Actual results #

```
level=WARN msg="upgrade failed" name=ceph error="OperatorConfig.csi.ceph.io \"ceph-csi-operator-config\" is invalid: [spec.driverSpecDefaults.enableMetadata: Invalid value: \"null\": spec.driverSpecDefaults.enableMetadata in body must be of type boolean: \"null\", <nil>: Invalid value: null: some validation rules were not checked because the object was invalid; correct the existing errors to complete validation]"
Error: UPGRADE FAILED: OperatorConfig.csi.ceph.io "ceph-csi-operator-config" is invalid: [spec.driverSpecDefaults.enableMetadata: Invalid value: "null": spec.driverSpecDefaults.enableMetadata in body must be of type boolean: "null", <nil>: Invalid value: null: some validation rules were not checked because the object was invalid; correct the existing errors to complete validation]
```

# Expected behavior #

Chart should install (presumably with the default enableMetadata of false).

It appears the chart template lacks a " | default false" or similar for enableMetadata.

iPraveenParihar:

The enableMetadata default was removed from values.yaml when the field was deprecated, but the templates still referenced it unconditionally. This caused helm install/upgrade to render enableMetadata as null, which the CRD rejects as an invalid boolean value.

Wrap the field with a nil check so it is omitted when not set (new installs) but still rendered when explicitly provided (upgrades from older chart versions).

Fixes: #473

---
_FYI: Future release of ceph-csi will remove the flag completely_
https://github.com/ceph/ceph-csi/blob/38c69b5dbecf924dfa83c43eecc469e705f03601/cmd/cephcsi.go#L215-L222

**Checklist:**

### DIFF
--- a/deploy/charts/ceph-csi-drivers/templates/driver.yaml
+++ b/deploy/charts/ceph-csi-drivers/templates/driver.yaml
@@ -31,7 +31,9 @@ spec:
   {{- if $driver.clusterName }}
   clusterName: {{ $driver.clusterName }}
   {{- end }}
+  {{- if ne $driver.enableMetadata nil }}
   enableMetadata: {{ $driver.enableMetadata }}
+  {{- end }}
   enableFencing: {{ $driver.enableFencing }}
   grpcTimeout: {{ $driver.grpcTimeout }}
   snapshotPolicy: {{ $driver.snapshotPolicy }}
--- a/deploy/charts/ceph-csi-drivers/templates/operatorConfig.yaml
+++ b/deploy/charts/ceph-csi-drivers/templates/operatorConfig.yaml
@@ -35,7 +35,9 @@ spec:
     {{- if $config.driverSpecDefaults.clusterName }}
     clusterName: {{ $config.driverSpecDefaults.clusterName }}
     {{- end }}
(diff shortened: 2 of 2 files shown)


Madhu-1 (approved):

Madhu-1:

PR #489, opened 2026-06-03
add design doc for ClientProfileReplication

rewantsoni:

Add the design doc related to the new ClientProfileReplication CR

- https://github.com/csi-addons/kubernetes-csi-addons/pull/993
- https://github.com/csi-addons/spec/pull/79
- https://github.com/ceph/ceph-csi/pull/6316

**Checklist:**

* [x] [Pending release notes](https://github.com/ceph/ceph-csi-operator/blob/main/PendingReleaseNotes.md) updated with breaking and/or notable changes for the next major release.
* [x] Unit tests have been added, if necessary.
* [x] Integration tests have been added, if necessary.

### DIFF
--- a/docs/design/replication.md
+++ b/docs/design/replication.md
@@ -1,0 +1,462 @@
+
+# ClientProfileReplication
+
+The ClientProfileReplication CR defines the replication destination mapping
+for backup and disaster recovery scenarios. It enables the CSI driver to determine the
+destination volume identifiers when client profile is different across cluster and pools
+have different IDs across mirrored Ceph clusters.
+
+## Purpose
+- Provides destination client profile and pool mapping for the `GetReplicationDestinationInfo` RPC
+- Enables DR/Backup orchestrators to discover correct destination volume IDs during failover
+- ClientProfileReplication replaces the replication destination discovery function of ClientProfileMapping. However, ClientProfileMapping must be retained as long as PersistentVolumes with stale cluster IDs exist in the cluster. Removing ClientProfileMapping prematurely will break all CSI operations on those volumes. Over time, as old PVs are deleted and replaced with PVs created through GetReplicationDestinationInfo (which have correct volume handles), ClientProfileMapping becomes unnecessary and can be safely removed.
+
+## Design Rationale
+
+1. **Pool name mapping**: Pool names are consistent across mirrored Ceph clusters,
+   while pool IDs can differ. Mapping by name is more intuitive and maintainable.
+   **Important**: Pool names must remain constant across clusters. If pools are
+   renamed on either cluster, the corresponding ClientProfileReplication CR must
+   be updated to reflect the new pool names.
+
+2. **Single destination**: The `GetReplicationDestinationInfo` RPC cannot receive
+   selection parameters, therefore only one destination per local client profile
+   is supported.
+
+3. **Operator translation**: The operator translates this CR into the
+   `replicationDestination` field in the ceph-csi-config ConfigMap, which the CSI
+   driver reads at runtime.
+
+## Constraints and Validation
+
+1. **Unique active client profiles**: Only one ClientProfileReplication CR may be in
+   `Ready` state for a given `localClientProfile`. If multiple CRs reference the same
+   `localClientProfile`:
+   - The oldest CR (by creation timestamp) is accepted and marked `Ready`
+   - All other CRs are rejected and marked `Rejected` with a descriptive message
+
+2. **Client profile dependency**: A ClientProfileReplication CR requires a
+   corresponding ClientProfile CR with a matching name.
+
+   - **Missing ClientProfile**: If a ClientProfileReplication references a
+     `localClientProfile` that doesn't exist, the CR is marked `Rejected` with message:
+     `"rejected: ClientProfile '<name>' not found"`
+
+   - **Deletion protection**: The ClientProfile controller adds a finalizer to
+     ClientProfile CR. If deletion is attempted, the finalizer blocks it until all referencing
+     ClientProfileReplication CRs are deleted first. This prevents breaking
+     replication configurations.
+
+## API Definition
+
+```go
+// ClientProfileReplicationSpec defines the desired replication destination mapping
+type ClientProfileReplicationSpec struct {
+    // LocalClientProfile is the name of the local ClientProfile CR
+    // +kubebuilder:validation:Required
+    LocalClientProfile string `json:"localClientProfile"`
+    
+    // RemoteClientProfile is the name of the remote cluster's client profile
+    // +kubebuilder:validation:Required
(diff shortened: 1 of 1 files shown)


docs/design/operator.md
```diff
@@ -384,6 +384,98 @@ spec:
...
+
+The ReplicationDestinationConfig CR defines the replication destination mapping
+for backup and disaster recovery scenarios. It enables the CSI driver to determine the
+destination volume identifiers when client profile is different across cluster and pools
+have different IDs across mirrored Ceph clusters.
+
+#### Purpose
+- Provides destination client profile and pool mapping for the `GetReplicationDestinationInfo` RPC
+- Enables DR/Backup orchestrators to discover correct destination volume IDs during failover
+- Serves as a replacement for ClientProfileMapping CR for more reliable replication.
```
iPraveenParihar: When you say replacement, will the ClientProfileMapping CRD deprecate?

docs/design/operator.md
```diff
@@ -384,6 +384,98 @@ spec:
     - [2, 1]
 ```
 
+### ReplicationDestinationConfig
+
+The ReplicationDestinationConfig CR defines the replication destination mapping
```
iPraveenParihar: Why do we need new CRD? The Purpose section defines what it does and not why it needs to be its own CRD?

I would argue to add it in ClientProfile CR since it represents profile's config and replication info (where the profile's data replicate to) should be property of ClientProfile?

Could you add justification on above^

docs/design/operator.md
```diff
@@ -384,6 +384,98 @@ spec:
...
+
+1. **Pool name mapping**: Pool names are consistent across mirrored Ceph clusters,
+   while pool IDs can differ. Mapping by name is more intuitive and maintainable.
+
+2. **Single destination**: The `GetReplicationDestinationInfo` RPC cannot receive
+   selection parameters, therefore only one destination per local client profile
+   is supported.
+
+3. **Operator translation**: The operator translates this CR into the
+   `replicationDestination` field in the csi-config ConfigMap, which the CSI
```
iPraveenParihar:
```suggestion
   `replicationDestination` field in the ceph-csi-config ConfigMap, which the CSI
```

iPraveenParihar: and everywhere else

iPraveenParihar (commented):
@rewantsoni Need clarification.

docs/design/operator.md
rewantsoni: Eventually, we can deprecate it but we cannot remove it since there is no way to know if the user moved to using the new ReplicationDestinationInfo CR

docs/design/operator.md
rewantsoni: 1. Theoritically can be multiple ReplicationDestinations for a single client profile, we shouldn't restrict this
2. I don't think we should bloat up the ClientProfile CR, they are related but they shouldn't be a part of the same API

docs/design/operator.md
```diff
@@ -384,6 +384,98 @@ spec:
     - [2, 1]
 ```
 
+### ReplicationDestinationConfig
```
Madhu-1: @rewantsoni let have this changes in a new file.

docs/design/operator.md
Madhu-1: replacement is not the right word as we are not replacing it in first go but eventually 1 will get deprecated and removed , something like below can be added

```
ReplicationDestinationConfig replaces the replication destination discovery function of ClientProfileMapping. However, ClientProfileMapping must be retained as long as PersistentVolumes with stale cluster IDs exist in the cluster. Removing ClientProfileMapping prematurely will break all CSI operations on those volumes. Over time, as old PVs are deleted and replaced with PVs created through GetReplicationDestinationInfo (which have correct volume handles), ClientProfileMapping becomes unnecessary and can be safely removed.
```

docs/design/operator.md
Madhu-1: How about naming it as ClientProfileReplication? as its point to the Replication details of the clientProfile

docs/design/operator.md
```diff
@@ -384,6 +384,98 @@ spec:
...
+  name: primary-to-secondary
+  namespace: <operator-namespace>
+spec:
+  localClientProfile: primary-cluster
+  remoteClientProfile: secondary-cluster
+  rbd:
+    remotePoolMapping:
+      # Pool "rbd" has ID 1 on primary, ID 5 on secondary
+      - poolName: rbd
+        poolID: "5"
```
Madhu-1: this is already under poolMapping, can this just be name and remoteID for more descriptive?

docs/design/operator.md
```diff
@@ -384,6 +384,98 @@ spec:
...
+kind: ReplicationDestinationConfig
+apiVersion: csi.ceph.io/v1
+metadata:
+  name: primary-to-secondary
+  namespace: <operator-namespace>
+spec:
+  localClientProfile: primary-cluster
+  remoteClientProfile: secondary-cluster
+  rbd:
+    remotePoolMapping:
```
Madhu-1: can this be poolMapping as the CR itself is for Replication?

Madhu-1 (commented):
Couple of comments when i look at this as standalone design doc.

* The design doc should also point to the consumption model on how DR will update the volume Handles
* Migration path for existing clusters, Need to cover what happens on the upgraded cluster where we have PVC with old volumeHandle
* CO-existance behaviour of both Replicationdesitination and client profile mapping
* What happens if 2 ReplicationDestinationConfig contains same localClientProfile?
* What happens if ReplicationDestinationConfig contains same localClientProfile but there is no clientProfile with the name yet?
* Missing details of which configmap key will be updataed for this new CRD
* Controller high level design, How 2 controllers will coordinates as they need to update the same key contents in the same configmap
* How we are ensureing the tight couping between the new CR and the clientProfile. if the clientProfile CR is deleted the entries in configmap will also gets
deleted even though we have ReplicationDestinationConfig
* Add a statement that poolNames are same across the clusters
* Add CRD defination as part of design
* Need to call out that design considered pool names as constamtt if ever pools are renamed. the corresponding details need to be updated on the CR as well.

docs/design/replication.md
```diff
@@ -0,0 +1,355 @@
...
+2. Look up ClientProfileReplication CRs using the index:
+   ```go
+   var replicationList ClientProfileReplicationList
+   err := r.List(ctx, &replicationList, 
+       client.InNamespace(clientProfile.Namespace),
+       client.MatchingFields{"spec.localClientProfile": clientProfile.Name})
+   ```
+3. **Validation**:
+   - If `len(replicationList.Items) > 1`, fail reconciliation and update ClientProfile
+     status with error:
```
Madhu-1: we also need to have a status updated on the ClientProfileReplication as well

Madhu-1 (approved):

PR #509, opened 2026-06-19
fix(helm): only render CephConnection readAffinity when crushLocationLabels is set

### LINKED ISSUE #508: CephConnection template renders null crushLocationLabels when read affinity is unset
# Describe the bug #

The `ceph-csi-drivers` chart always renders `spec.readAffinity.crushLocationLabels`
in `templates/cephconnection.yaml`, even when no CRUSH location labels are set.
With an empty/unset list the key is emitted with no value, which YAML treats as
`null`. Since the `CephConnection` CRD requires `crushLocationLabels` to be a
non-empty array when present, the resource is rejected and cannot be created.
This makes it impossible to deploy a `CephConnection` without enabling read
affinity, although the field is meant to be optional.

# Environment details #

- Image/version of ceph-csi-operator : ceph-csi-drivers chart v1.0.1

# Steps to reproduce #

1. Use the `ceph-csi-drivers` chart with a `cephConnections` entry that does not
   set `crushLocationLabels` (or sets it to `[]` / `{}`).
2. Install/sync the chart so the `CephConnection` CR is applied.
3. See the validation error.

# Actual results #

The `CephConnection` is rejected:
```
CephConnection.csi.ceph.io "ceph-cluster" is invalid:
spec.readAffinity.crushLocationLabels: Invalid value: "null":
spec.readAffinity.crushLocationLabels in body must be of type array: "null"
```

The only way to make it pass is to provide at least one label, which forces read
affinity on.

# Expected behavior #

When `crushLocationLabels` is empty/unset, the `readAffinity` block should be
omitted entirely so the optional field is simply not set, and the
`CephConnection` should be created successfully.

# Logs #

(see validation error above)

# Additional context #

A related issue already exists for this behavior. I'm opening this for tracking
and will provide the fix in an upcoming PR. https://github.com/ceph/ceph-csi-operator/issues/382

NicolasBoulard:

The `cephconnection.yaml` template in the `ceph-csi-drivers` chart always
renders the `spec.readAffinity.crushLocationLabels` key, regardless of whether
any CRUSH location labels are configured:

```yaml
  readAffinity:
    crushLocationLabels:
    {{- range $cephConnection.crushLocationLabels }}
    - {{ . }}
    {{- end }}
```

When `crushLocationLabels` is empty or unset (which is the chart default), the
`range` produces nothing and the rendered output becomes:

```yaml
  readAffinity:
    crushLocationLabels:
```

YAML parses that valueless key as `null`. The `CephConnection` CRD defines
`crushLocationLabels` as a required, non-empty array of strings inside the
otherwise-optional `readAffinity` block (`ReadAffinitySpec.CrushLocationLabels`
is `Optional`-on-`readAffinity` but `Required` + `MinItems=1` once present), so
the `null` value fails validation and the resource is rejected:
```
CephConnection.csi.ceph.io "ceph-cluster" is invalid:
spec.readAffinity.crushLocationLabels: Invalid value: "null":
spec.readAffinity.crushLocationLabels in body must be of type array: "null"
```

This makes it impossible to deploy a `CephConnection` without enabling read
affinity, even though both `readAffinity` and the field are meant to be
optional.

This PR wraps the whole `readAffinity` block in a `with`, so it is only rendered
when at least one CRUSH location label is provided:

```yaml
  rbdMirrorDaemonCount: {{ $cephConnection.rbdMirrorDaemonCount }}
  {{- with $cephConnection.crushLocationLabels }}
  readAffinity:
    crushLocationLabels:
    {{- range . }}
    - {{ . }}
    {{- end }}
  {{- end }}
```

With this change:
- no labels configured : the `readAffinity` block is omitted entirely and the
  CRD accepts the resource (field is optional);
- labels configured : output is unchanged from before.

* The change is backward compatible. Users who set `crushLocationLabels` get
  exactly the same rendered output as today. Users who left it empty previously
  produced an invalid `CephConnection`; they now get a valid one with read
  affinity disabled, which matches the documented default (`{}`).
* No values-schema change is required; `cephConnections[].crushLocationLabels`
  keeps its current default.

## Related issues ##

Fixes: #508

## Future concerns ##

**Checklist:**

  guidelines in the [developer
* [ ] [Pending release

### DIFF
--- a/deploy/charts/ceph-csi-drivers/templates/cephConnection.yaml
+++ b/deploy/charts/ceph-csi-drivers/templates/cephConnection.yaml
@@ -12,10 +12,12 @@ spec:
   - {{ . }}
   {{- end }}
   rbdMirrorDaemonCount: {{ $cephConnection.rbdMirrorDaemonCount }}
+  {{- with $cephConnection.crushLocationLabels }}
   readAffinity:
     crushLocationLabels:
-    {{- range $cephConnection.crushLocationLabels }}
+    {{- range . }}
     - {{ . }}
     {{- end }}
+  {{- end }}
 {{- end }}
 {{- end }}
--- a/deploy/charts/ceph-csi-drivers/values.yaml
+++ b/deploy/charts/ceph-csi-drivers/values.yaml
@@ -6,8 +6,9 @@ cephConnections:
     monitors: {}
     # -- Number of RBD mirror daemons (default: 1)
     rbdMirrorDaemonCount: 1
-    # -- Labels to be used for CRUSH location selection (default: {})
-    crushLocationLabels: {}
+    # -- Labels to be used for CRUSH location selection (default: [])
+    # Leave empty to disable read affinity. When set, it must be a non-empty list of node label keys, e.g. ["topology.kubernetes.io/zone"].
+    crushLocationLabels: []
 # Configuration for Client Profiles
 clientProfiles:
   - # -- Name of the client profile (default: "")
--- a/docs/helm-charts/drivers-chart.md
+++ b/docs/helm-charts/drivers-chart.md
@@ -65,7 +65,7 @@ The following table lists the configurable parameters of the ceph-csi-drivers ch
 
 | Parameter | Description | Default |
 |-----------|-------------|---------|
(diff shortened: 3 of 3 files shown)


Madhu-1 (approved):

Madhu-1:

PR #517, opened 2026-06-23
controller: set replicas if not set

### LINKED ISSUE #499: Only deploy one controller plugin replica in single-node clusters
# Describe the feature you'd like to have #

Production clusters should use two controllers to ensure redundancy. 
Single-node clusters for testing can only start one controller. If two controllers attempt to start, one of them will be stuck in pending due to the pod anti-affinity.

As a convenience during testing on single-node clusters, it will be very helpful if the csi operator restricted the controllers internally to only start a single instance even if two are in the CR. Of course, this is not strictly respecting the desired state and there is a simple workaround to change the replicas to 1, it is just a convenience. When Rook managed the csi settings, Rook implemented this convenience [here](https://github.com/rook/rook/blob/release-1.19/pkg/operator/ceph/csi/csi.go#L269-L273). Now with the CSI operator, we are hoping to restore this convenience.

## Who is the end user and what is the use case where this feature will be valuable? ##
  
During development, this would be very helpful for minikube testing.

# How will we know we have a good solution? (acceptance criteria) #

Minikube testing is smooth.

Madhu-1:
If the replica count is not set in the CR set the replica count to 2 if the number of nodes is > 1 or if its single node cluster set the replica count to 1.

fixes: #499

### DIFF
--- a/internal/controller/driver_controller.go
+++ b/internal/controller/driver_controller.go
@@ -60,6 +60,7 @@ import (
 //+kubebuilder:rbac:groups=storage.k8s.io,resources=csidrivers,verbs=get;list;watch;create;update;patch;delete
 //+kubebuilder:rbac:groups=apps,resources=deployments,verbs=get;list;watch;create;update;patch;delete
 //+kubebuilder:rbac:groups=apps,resources=daemonsets,verbs=get;list;watch;create;update;patch;delete
+//+kubebuilder:rbac:groups="",resources=nodes,verbs=list;watch
 //+kubebuilder:rbac:groups="",resources=services,verbs=get;list;watch;create;update;patch;delete
 
 type DriverType string
@@ -566,8 +567,10 @@ func (r *driverReconcile) reconcileControllerPluginDeployment() error {
 		nodePluginSpec := cmp.Or(r.driver.Spec.NodePlugin, &csiv1.NodePluginSpec{})
 		topology := r.isRbdDriver() && nodePluginSpec.Topology != nil
 
+		replicas := r.getControllerPluginReplicas(log, pluginSpec.Replicas)
+
 		deploy.Spec = appsv1.DeploymentSpec{
-			Replicas: pluginSpec.Replicas,
+			Replicas: replicas,
 			Selector: &appSelector,
 			Strategy: ptr.Deref(pluginSpec.DeploymentStrategy, defaultDeploymentStrategy),
 			Template: corev1.PodTemplateSpec{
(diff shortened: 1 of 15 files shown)


internal/controller/driver_controller.go
```diff
@@ -1667,6 +1670,31 @@ func (r *driverReconcile) generateServiceName(suffix string) string {
 	return re.ReplaceAllString(name, "-")
 }
 
+func (r *driverReconcile) getControllerPluginReplicas(
```
travisn, line 1673: unit tests for this method?

internal/controller/driver_controller.go
```diff
@@ -1667,6 +1670,31 @@ func (r *driverReconcile) generateServiceName(suffix string) string {
...
+	nodeList := &corev1.NodeList{}
+	if err := r.List(r.ctx, nodeList); err != nil {
+		log.Error(err, "Failed to list nodes for replica calculation")
+		return &replicas
+	}
+
+	nodeCount := len(nodeList.Items)
+	if nodeCount > 0 && nodeCount < int(replicas) {
+		replicas = 1
+		log.Info("Capping controller replicas to node count",
```
travisn, line 1691: From #499 I had understood that if the CR value was set, it would always be honored. This looks different to always cap the number of replicas to the number of nodes even if the CR value is set. This approach is effectively the behavior that we had in Rook previously which is likely simplest to implement, I just want to be clear on the intention.

Madhu-1: it still the same https://github.com/ceph/ceph-csi-operator/pull/517/changes#diff-3070e6d2f0c819fa25f1a293866ae019d29cdfe2d5e4ed4375d295cf2cdfd6f2R1677

1. If replicas are set return it
2. If empty and if node count is 1 return 1
3. If empty and node count is not 1 return 2

Madhu-1:

travisn:
In Rook testing with v1.0.3, the changes from this PR resulted in the following unexpected behavior:
- The default value in the CRD is 2. The absence of the setting in the CR is interpreted as 2, and we never hit the code path that checks for nodes. The CR setting is always applied, and never defaults to 1 for a single node. 
- The `replicas` value set in OperatorConfig CR does not apply to the driver. If setting `replicas: 1` in the OperatorConfig, two replicas are still started for each driver. The only way to set `replicas: 1` currently is on each individual driver CR. The OperatorConfig `replicas` does not seem to apply.

PR #194, opened 2025-01-13
add a hostNetwork setting to the  Driver and ControllerPlugin   Specs

obnoxxx:
This bool setting can be added to the controllerPlugin section of the Driver Spec.

It will be propagated to all controller plugin pods.

This implements the following design:

https://github.com/ceph/ceph-csi-operator/blob/main/docs/design/hostNetwork.md

see issue #157  for background and context.

# Describe what this PR does #

This work in progress is a draft PR for the implementation of host Networking for controller plugin pods  as described in https://github.com/ceph/ceph-csi-operator/blob/main/docs/design/hostNetwork.md

## Is there anything that requires special attention ##

This needs thorough testing.

Is the change backward compatible?

should be but this  needs testing.

Are there concerns around backward compatibility?

none right now

For example:

## Related issues ##

Fixes: #157 

## Future concerns ##

none. 

**Checklist:**

* [x] **Commit Message Formatting**: Commit titles and messages follow
  guidelines in the [developer
  guide](https://github.com/ceph/ceph-csi-operator/blob/main/docs/development-guide.md#commit-messages).
* [x] Reviewed the developer guide on [Submitting a Pull
  Request](https://github.com/ceph/ceph-csi-operator/blob/main/docs/development-guide.md#development-workflow)
* [ ] [Pending release
  notes](https://github.com/ceph/ceph-csi-operator/blob/main/PendingReleaseNotes.md)
  updated with breaking and/or notable changes for the next major release.
* [ ] Documentation has been updated, if necessary.
* [x] Unit tests have been added, if necessary.
* [ ] Integration tests have been added, if necessary.

PR #520, opened 2026-06-24
feat: make nodePlugin hostNetwork configurable

### LINKED ISSUE #521: nodePlugin hostNetwork:true hardcoded prevents Istio mTLS sidecar injection
## Summary

The CSI node plugin DaemonSet has `hostNetwork: true` hardcoded in the
controller. This makes it impossible to run the node plugin inside an Istio
service mesh that enforces mutual TLS (`PeerAuthentication: STRICT`), because
Istio cannot inject its sidecar proxy into pods with `hostNetwork: true`.

## Background

In environments with strict mTLS requirements — such as SAP's internal security
standards — all in-cluster service-to-service communication must be encrypted
and mutually authenticated at the transport layer via Istio. This applies to the
node plugin's connection to vault-like secret backends (e.g. a key-manager
service used for LUKS volume encryption).

The node plugin needs to authenticate to the key-manager at volume mount time to
retrieve the LUKS passphrase. With `hostNetwork: true`, Istio cannot inject a
sidecar, so the pod has no mTLS identity. The key-manager's
`PeerAuthentication: STRICT` policy then rejects the connection with a TCP reset,
causing all encrypted volume mounts to fail.

## Impact

- Any deployment running Istio with `PeerAuthentication: STRICT` on vault-like
  backends (HashiCorp Vault, custom key managers, etc.) cannot use RBD volume
  encryption
- Workaround requires weakening the mTLS policy to `PERMISSIVE` on the
  key-manager service, which violates the security standard

## Validation

We manually patched the live DaemonSet to `hostNetwork: false` in a production
Rook-Ceph v1.19.6 cluster and confirmed:

- Istio sidecar is injected on pod restart ✓
- mTLS is established with the key-manager ✓
- Encrypted RBD volumes mount successfully end-to-end ✓
- `hostPID: true` continues to work (required for LUKS/cryptsetup nsenter) ✓
- Ceph mon connectivity is unaffected (cluster DNS resolves mons correctly) ✓

`hostNetwork: true` is not a hard requirement for RBD functionality. It was
historically set as a default for mon FQDN resolution but works fine without it
when cluster DNS is available, which is the case in all standard Kubernetes
deployments.

## Proposed Fix

Add `hostNetwork *bool` to `NodePluginSpec` with default `true` (backward
compatible), mirroring the existing configurable `hostNetwork` field already
present on `ControllerPluginSpec` (added in #194).

PR: #520

## Notes

- `ControllerPluginSpec.hostNetwork` is already configurable since #194 —
  the node plugin is the remaining gap
- The change is a single-field addition with a safe default; no existing
  deployments are affected unless they explicitly opt in

sergeykuperman:
## Problem

The CSI node plugin DaemonSet has `hostNetwork: true` hardcoded in the
controller. This prevents Istio sidecar injection, making it impossible
to use the node plugin in service meshes that enforce mTLS
(`PeerAuthentication: STRICT`).

The controller plugin already has a configurable `hostNetwork` field on
`ControllerPluginSpec` (added in #194). The node plugin has no equivalent.

Closes #521

## Solution

Add an optional `hostNetwork` field to `NodePluginSpec`, mirroring the
existing field on `ControllerPluginSpec`.

- **Default: `true`** — fully backward compatible, no behavior change for
  existing users
- When set to `false`, DNS policy is automatically switched from
  `ClusterFirstWithHostNet` to `ClusterFirst`
- `hostPID: true` is preserved regardless (required for LUKS/cryptsetup)

## Validation

Tested in a production Rook-Ceph v1.19.6 cluster running Istio with
`PeerAuthentication: STRICT` on the key-manager service:

1. Patched the live DaemonSet to `hostNetwork: false`
2. Confirmed Istio sidecar was injected on pod restart
3. Confirmed mTLS was established with the key-manager (LUKS encryption service)
4. Confirmed encrypted RBD volume mounts succeeded end-to-end
5. All existing controller tests pass (`make test`)

## Changes

- `api/v1/driver_types.go` — add `HostNetwork *bool` to `NodePluginSpec`
- `api/v1/zz_generated.deepcopy.go` — regenerated
- `config/crd/bases/csi.ceph.io_drivers.yaml` — regenerated CRD
- `internal/controller/driver_controller.go` — use `ptr.Deref(pluginSpec.HostNetwork, true)` and conditional DNS policy
- `vendor/` — updated vendored copies accordingly

Signed-off-by: Sergey Kuperman <sergey.kuperman@sap.com>


Madhu-1:
@sergeykuperman its not possible to run the nodeplugin on the pod network because if the nodeplugin restarts we will lose all the mounts and all the I/O is going to hang because of that we run the nodeplugin always on the host networking

sergeykuperman:
we are currently running version 1.16.6 and we set the network to pod network and everything works. why exactly was this functionality broken in newest versions ? we need MTLS between the plugin and vault-like api for provisioning LUKS encryption keys.. and istio wont inject it's sidecar into pods on the host network

Madhu-1:
> we are currently running version 1.16.6 and we set the network to pod network and everything works. why exactly was this functionality broken in newest versions ? we need MTLS between the plugin and vault-like api for provisioning LUKS encryption keys.. and istio wont inject it's sidecar into pods on the host network

what is 1.16.6 version are you referring here? did you try to restart the nodeplugin and the application pods are still able to do I/O?

sergeykuperman:
> > we are currently running version 1.16.6 and we set the network to pod network and everything works. why exactly was this functionality broken in newest versions ? we need MTLS between the plugin and vault-like api for provisioning LUKS encryption keys.. and istio wont inject it's sidecar into pods on the host network
> 
> what is 1.16.6 version are you referring here? did you try to restart the nodeplugin and the application pods are still able to do I/O?

i mean rook-ceph chart:
where it was part of rook itself:
https://github.com/rook/rook/tree/release-1.16/deploy/charts/rook-ceph
and yes pods continue to function, and are able to terminate successfully with hostNetwork: false in this version

Madhu-1:
> > > we are currently running version 1.16.6 and we set the network to pod network and everything works. why exactly was this functionality broken in newest versions ? we need MTLS between the plugin and vault-like api for provisioning LUKS encryption keys.. and istio wont inject it's sidecar into pods on the host network
> > 
> > 
> > what is 1.16.6 version are you referring here? did you try to restart the nodeplugin and the application pods are still able to do I/O?
> 
> i mean rook-ceph chart: where it was part of rook itself: https://github.com/rook/rook/tree/release-1.16/deploy/charts/rook-ceph and yes pods continue to function, and are able to terminate successfully with hostNetwork: false in this version

It looks like you are running very old version of Rook, Rook removed the support for running nodeplugin in pod network very long back, In Rook CSI always runs on host networking.

sergeykuperman:
Understood — makes sense that hostNetwork is required to avoid disrupting existing mounts on restart. Since the pod runs on the host network and sidecar injection isn't possible, the only way to implement mTLS between the node 
  plugin and a vault-like backend would be at the application level (e.g. TLS client certificates for the vault HTTP client). Are there any plans to support that in the ceph-csi vault KMS driver? Our environment requires mTLS for
  all in-cluster service-to-service communication, and currently we have to run the connection with PeerAuthentication: PERMISSIVE on the relevant port as a workaround

Madhu-1:
> Understood — makes sense that hostNetwork is required to avoid disrupting existing mounts on restart. Since the pod runs on the host network and sidecar injection isn't possible, the only way to implement mTLS between the node plugin and a vault-like backend would be at the application level (e.g. TLS client certificates for the vault HTTP client). Are there any plans to support that in the ceph-csi vault KMS driver? Our environment requires mTLS for all in-cluster service-to-service communication, and currently we have to run the connection with PeerAuthentication: PERMISSIVE on the relevant port as a workaround

Thanks, that would be a different discussion, please file a feature request on the ceph-csi repository describing your KMS setup and the TLS requirements.

PR #529, opened 2026-07-01
ci: use tag name instead of GIT_REF

Madhu-1:
The stable release publish was triggered by the release: published event, which intermittently delivers an empty github.ref (actions/runner#2788). using `github.event.release.tag_name` instead, as this is the preferred method for referencing the tag name of a release.


Madhu-1:

PR #538, opened 2026-07-06
remove the default replicas from validation

### LINKED ISSUE #532: due to default replica count, on single node the csi deployment is not created with replica 1
In Rook testing with v1.0.3, the changes from this PR resulted in the following unexpected behavior:
- The default value in the CRD is 2. The absence of the setting in the CR is interpreted as 2, and we never hit the code path that checks for nodes. The CR setting is always applied, and never defaults to 1 for a single node. 
- The `replicas` value set in OperatorConfig CR does not apply to the driver. If setting `replicas: 1` in the OperatorConfig, two replicas are still started for each driver. The only way to set `replicas: 1` currently is on each individual driver CR. The OperatorConfig `replicas` does not seem to apply.

_Originally posted by @travisn in https://github.com/ceph/ceph-csi-operator/issues/517#issuecomment-4859175129_

Madhu-1:
Removing the default replica count from the CRD to ensure that we can still allow the user to deploy single controller plugin deployment on the single node cluster.

fixes: #532

Scenario | Expected Replicas | Actual Replicas | Result
-- | -- | -- | --
Driver CR without replicas, no OperatorConfig | 1 (node-count capped) | 1 | Pass
Driver CR without replicas, OperatorConfig replicas: 2 | 2 (propagated from OperatorConfig) | 2 | Pass
Driver CR with explicit replicas: 3 | 3 (user override wins) | 3 | Pass

### DIFF
--- a/api/v1/driver_types.go
+++ b/api/v1/driver_types.go
@@ -229,12 +229,12 @@ type ControllerPluginSpec struct {
 	//+kubebuilder:validation:Optional
 	DeploymentStrategy *appsv1.DeploymentStrategy `json:"deploymentStrategy,omitempty"`
 
-	// Set replicas for controller plugin's deployment. Defaults to 2.
-	// On single-node clusters, the operator automatically caps the replica
-	// count to the number of available nodes when this field is not set.
+	// Set replicas for controller plugin's deployment.
+	// If not set, defaults to 2 or to the number of available nodes
+	// on single-node clusters. When OperatorConfig specifies a replica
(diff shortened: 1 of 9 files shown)


Madhu-1:
@travisn PTAL all the cases should work , the only issue was with default value in validation

travisn:
LGTM, agreed that removing the default crd value is expected to fix the reported issue, thanks!

Madhu-1:

PR #544, opened 2026-07-13
Add NetworkPolicy for operator and driver pods

Rakshith-R:
## Summary

- Add a static NetworkPolicy for the operator pod that denies all ingress
  and allows open egress (API server access). Included in all generated
  manifests and helm charts by default.
- Add code-generated NetworkPolicies for controller-plugin and csi-addons
  nodeplugin pods as part of the driver reconciliation loop.
- Controller-plugin NP:
  - Skipped when `hostNetwork: true` (NP exempt)
  - csi-addons gRPC ingress rule added only when `DeployCsiAddons: true`
  - Snapshot-metadata port 50051 ingress added only for RBD with TLS
    configured (`tls-key` volume present)
  - Deny-all ingress (empty rules) when no features enabled —
    defense-in-depth
- csi-addons nodeplugin NP:
  - Ingress from csi-addons controller-manager on port 9071
  - Explicitly deleted when `DeployCsiAddons` toggled false
  - Driver CR as controller owner for watch triggers and GC
- Node-plugin DaemonSet is exempt (`hostNetwork: true`).
- Fix `yaml-upgrade-test` to use `kubectl apply --server-side` instead of
  `kubectl replace` (which fails when upgrading adds new resource types).

### Why unrestricted egress (`- {}`)

All policies use open egress instead of restricting to specific IPs because
the API server `ClusterIP` is not portable:

1. **Service CIDR varies per cluster** — `10.96.0.1` (kubeadm),
   `172.30.0.1` (OpenShift), custom ranges on managed clouds.
   Unknown at manifest authoring time.
2. **CNI `ipBlock` behavior is inconsistent** — OVN-Kubernetes applies
   NetworkPolicy before or after DNAT depending on traffic direction,
   making `ipBlock`-based rules unreliable for service IPs
   ([OVN-K8s docs](https://docs.openshift.com/container-platform/latest/networking/ovn_kubernetes_network_provider/about-ovn-kubernetes.html)).
   Cilium has [similar quirks](https://github.com/cilium/cilium/issues/12277).
3. **HyperShift / non-standard ports** — HyperShift clusters expose the
   API server on non-6443 ports. Hardcoding port 6443 breaks these setups.
4. **Precedent** — OpenShift networking team guidance in
   [operator-marketplace PR #723](https://github.com/operator-framework/operator-marketplace/pull/723)
   uses open egress for the same reasons.
5. **Ceph cluster access** — Driver pods also need egress to the Ceph
   monitor/OSD network, which varies per deployment.

### References

- [CSI snapshot-metadata sidecar](https://kubernetes-csi.github.io/docs/external-snapshot-metadata.html) — port 50051 clients are backup apps (Velero)
- [kubernetes-csi/external-snapshot-metadata](https://github.com/kubernetes-csi/external-snapshot-metadata)
- [operator-marketplace NP precedent](https://github.com/operator-framework/operator-marketplace/pull/723)
- [OVN-K8s NetworkPolicy + DNAT](https://docs.openshift.com/container-platform/latest/networking/ovn_kubernetes_network_provider/about-ovn-kubernetes.html)
- [Cilium ipBlock + service DNAT issue](https://github.com/cilium/cilium/issues/12277)
- Design doc: `docs/design/network-policy.md`

## Test plan

- [x] `make build` passes
- [x] `make test` passes (existing tests)
- [x] `make build-installer` includes NP in output
- [x] Deploy operator with NP, verify pod remains Running
- [x] Create Driver CR, verify controller-plugin NP created
- [x] Toggle `DeployCsiAddons: false`, verify nodeplugin NP deleted
- [x] PVC provision (RBD + CephFS) works with NPs applied
- [x] CSIAddonsNode CRs register (gRPC connectivity through NP)

Manually tested on live OCP 4.23 cluster — PVC provision,
CSIAddonsNode registration, operator reconcile all pass.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/PendingReleaseNotes.md
+++ b/PendingReleaseNotes.md
@@ -1,11 +1,9 @@
-# v1.0.0 Pending Release Notes
+# v1.1.0 Pending Release Notes
 
 ## Breaking Changes
 
 ## Features
 
-- Added `containerExtraArgs` field to `NodePluginSpec` and `ControllerPluginSpec` to allow passing custom arguments to CSI containers. The field accepts a map where the key is the container name (e.g., `csi-rbdplugin`, `csi-provisioner`, `driver-registrar`) and the value is a list of CLI arguments. This enables customization of container behavior without modifying default operator values. The field can be configured at both the OperatorConfig level (for defaults) and Driver level (for driver-specific overrides).
-- Fencing can now be enabled in the ceph-csi-drivers Helm chart either globally for all drivers or on a per-driver basis.
-- Added `enabled` field to log rotation configuration in the ceph-csi-drivers Helm chart. Users can now disable log rotation by setting `log.rotation.enabled: false` either globally in `operatorConfig.driverSpecDefaults` or per-driver in `drivers.<driver-type>`. This resolves Helm warnings when disabling file logging from parent charts.
-- Added support to set the priorityClass in the operator helm chart.
+- Added NetworkPolicies for the operator pod and CSI driver pods (controller-plugin, csi-addons nodeplugin). Included in all generated manifests by default. Driver pod NPs are created by the operator for every reconciled driver. Node-plugin pods are exempt (`hostNetwork: true`).
+
 ## NOTE
--- a/config/network-policy/operator-network-policy.yaml
+++ b/config/network-policy/operator-network-policy.yaml
@@ -1,0 +1,14 @@
+apiVersion: networking.k8s.io/v1
+kind: NetworkPolicy
+metadata:
+  name: ceph-csi-controller-manager
+spec:
+  podSelector:
+    matchLabels:
+      control-plane: ceph-csi-op-controller-manager
+  # Ingress is not specified — deny all inbound traffic.
+  egress:
+  - {}
+  policyTypes:
+  - Ingress
+  - Egress
--- a/docs/design/network-policy.md
+++ b/docs/design/network-policy.md
@@ -36,17 +36,17 @@ ceph-csi-operator and the CSI driver pods it manages.
      applications (Velero, etc.) can run in any namespace with any labels.
      See [CSI snapshot-metadata](https://kubernetes-csi.github.io/docs/external-snapshot-metadata.html).
 - **Egress:** Open. Required for API server and Ceph cluster access.
-- **Deleted** when no ingress rules apply (no csi-addons, no
-  snapshot-metadata).
+- When no ingress rules apply, the NP is kept with an empty ingress
+  list (deny-all ingress) for defense-in-depth.
 
 ### csi-addons Nodeplugin
 
 - **Ingress:** csi-addons controller-manager → port 9071 only.
 - **Egress:** Open.
 - **Lifecycle:** Created when `DeployCsiAddons` is true. Not created for
-  NFS drivers. The csi-addons DaemonSet is set as an owner reference —
-  when the DaemonSet is deleted (DeployCsiAddons toggled false), the NP
-  is garbage collected automatically.
+  NFS drivers. Explicitly deleted when `DeployCsiAddons` is toggled
+  false. The Driver CR is the controller owner reference for watch
+  triggers and garbage collection on Driver deletion.
 
 ### Node-Plugin DaemonSet
 
--- a/docs/design/network-policy.md
+++ b/docs/design/network-policy.md
@@ -1,0 +1,88 @@
+# NetworkPolicy for Ceph-CSI Operator and Driver Pods
+
+This document describes the NetworkPolicy implementation for the
+ceph-csi-operator and the CSI driver pods it manages.
+
+## Component Inventory
+
+| Component | Pod Labels | hostNetwork | NP Coverage |
+|---|---|---|---|
+| Operator pod | `control-plane: ceph-csi-op-controller-manager` | No | Static manifest |
+| Controller-plugin (per driver) | `app: {driver}-ctrlplugin` | Configurable | Code-generated |
+| csi-addons node agent (per driver) | `app: {driver}-nodeplugin-csi-addons` | No | Code-generated |
+| Node-plugin DaemonSet | `app: {driver}-nodeplugin` | Yes (hardcoded) | Exempt |
+
+## Policies
+
+### Operator Pod
+
+- **Ingress:** Deny all. No exposed ports (metrics disabled by default,
+  kubelet health probes bypass NetworkPolicy).
+- **Egress:** Open (`- {}`). Required for API server access. The API server
+  ClusterIP varies per cluster (10.96.0.1 on kubeadm, 172.30.0.1 on OpenShift)
+  and CNI ipBlock behavior is inconsistent across implementations, so
+  restricting egress to a specific IP is not portable.
+
+### Controller-Plugin
+
+- **Skipped** when `hostNetwork: true` (NP is exempt).
+- **Ingress** rules are conditional:
+  1. (When `DeployCsiAddons: true`) csi-addons controller-manager →
+     csi-addons gRPC port (9070 for RBD, 9080 for CephFS). Source
+     restricted to pods with labels `app.kubernetes.io/name: csi-addons`
+     and `control-plane: controller-manager` in any namespace.
+  2. (RBD only, when TLS volume `tls-key` is configured) Any pod →
+     snapshot-metadata gRPC port 50051. Open source because backup
+     applications (Velero, etc.) can run in any namespace with any labels.
+     See [CSI snapshot-metadata](https://kubernetes-csi.github.io/docs/external-snapshot-metadata.html).
+- **Egress:** Open. Required for API server and Ceph cluster access.
+- **Deleted** when no ingress rules apply (no csi-addons, no
+  snapshot-metadata).
+
+### csi-addons Nodeplugin
+
+- **Ingress:** csi-addons controller-manager → port 9071 only.
+- **Egress:** Open.
+- **Lifecycle:** Created when `DeployCsiAddons` is true. Not created for
+  NFS drivers. The csi-addons DaemonSet is set as an owner reference —
+  when the DaemonSet is deleted (DeployCsiAddons toggled false), the NP
+  is garbage collected automatically.
+
+### Node-Plugin DaemonSet
+
+No NetworkPolicy. Runs with `hostNetwork: true`, which makes it exempt
+from NetworkPolicy enforcement.
+
+## Inclusion
+
+### Static Manifest (Operator Pod)
+
+The operator pod NP is in `config/network-policy/` and referenced from
+`config/default/kustomization.yaml`. It is included in all generated
+manifests by default.
+
+### Code-Generated NPs (Driver Pods)
+
+Driver pod NPs are created unconditionally by the operator at runtime.
+The `networking.k8s.io/v1` API group has been stable since Kubernetes 1.7
+and is available on every supported cluster version.
+
+## API Server Egress Rationale
+
+Open egress (`- {}`) is used instead of restricting to a specific API server
+IP because:
+
+1. The API server ClusterIP varies per cluster and is unknown at manifest
+   authoring time.
+2. CNI implementations handle `ipBlock` inconsistently with DNAT (OVN-K8s
+   timing, Cilium quirks).
+3. HyperShift clusters may use non-standard API server ports.
+
+This follows the precedent set by
+[operator-marketplace PR #723](https://github.com/operator-framework/operator-marketplace/pull/723).
+
+## References
+
+- [CSI snapshot-metadata sidecar](https://kubernetes-csi.github.io/docs/external-snapshot-metadata.html)
+- [kubernetes-csi/external-snapshot-metadata](https://github.com/kubernetes-csi/external-snapshot-metadata)
+- [operator-marketplace NP precedent](https://github.com/operator-framework/operator-marketplace/pull/723)
--- a/internal/controller/driver_controller.go
+++ b/internal/controller/driver_controller.go
@@ -32,6 +32,7 @@ import (
 	"github.com/go-logr/logr"
 	appsv1 "k8s.io/api/apps/v1"
 	corev1 "k8s.io/api/core/v1"
+	networkingv1 "k8s.io/api/networking/v1"
 	storagev1 "k8s.io/api/storage/v1"
 	k8serrors "k8s.io/apimachinery/pkg/api/errors"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
@@ -62,6 +63,7 @@ import (
 //+kubebuilder:rbac:groups=apps,resources=daemonsets,verbs=get;list;watch;create;update;patch;delete
 //+kubebuilder:rbac:groups="",resources=nodes,verbs=list;watch
 //+kubebuilder:rbac:groups="",resources=services,verbs=get;list;watch;create;update;patch;delete
+//+kubebuilder:rbac:groups=networking.k8s.io,resources=networkpolicies,verbs=get;list;watch;create;update;patch;delete
 
 type DriverType string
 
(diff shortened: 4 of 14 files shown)


docs/design/network-policy.md
```diff
@@ -0,0 +1,82 @@
...
+  restricting egress to a specific IP is not portable.
+
+### Controller-Plugin
+
+- **Ingress:** Two rules:
+  1. csi-addons controller-manager → csi-addons gRPC port (9070 for RBD,
+     9080 for CephFS). Source restricted to pods with labels
+     `app.kubernetes.io/name: csi-addons` + `control-plane: controller-manager`
+     in any namespace.
+  2. (RBD only) Any pod → snapshot-metadata gRPC port 50051. Open source
```
Madhu-1: what happens when the controller runs on the host network?

internal/controller/driver_controller.go
```diff
@@ -1044,6 +1049,61 @@ func (r *driverReconcile) reconcileControllerPluginDeployment() error {
 	return err
 }
 
+func (r *driverReconcile) reconcileControllerPluginNetworkPolicy() error {
```
Madhu-1, line 1052: what happens when the controller plugin is running on host network which means csi-addons will also run on host network?

internal/controller/driver_controller.go
```diff
@@ -1260,6 +1320,54 @@ func (r *driverReconcile) reconcileNodePluginDaemonSetForCsiAddons() error {
 	return err
 }
 
+func (r *driverReconcile) reconcileNodeAgentNetworkPolicy() error {
```
Madhu-1: rename it to `reconcileNodePluginCsiAddonsNetworkPolicy`

internal/controller/driver_controller.go
```diff
@@ -1260,6 +1320,54 @@ func (r *driverReconcile) reconcileNodePluginDaemonSetForCsiAddons() error {
...
+
+	if !ptr.Deref(r.driver.Spec.DeployCsiAddons, false) || r.isNfsDriver() {
+		if err := r.Delete(r.ctx, np); client.IgnoreNotFound(err) != nil {
+			return err
+		}
+		return nil
+	}
+
+	log := r.log.WithValues("networkPolicy", np.Name)
+	log.Info("Reconciling csi-addons node agent network policy")
```
Madhu-1: there is nothing called node agent, we should log this as csiaddons for nodeplugin or similar

internal/controller/driver_controller.go
```diff
@@ -1260,6 +1320,54 @@ func (r *driverReconcile) reconcileNodePluginDaemonSetForCsiAddons() error {
 	return err
 }
 
+func (r *driverReconcile) reconcileNodeAgentNetworkPolicy() error {
+	np := &networkingv1.NetworkPolicy{}
+	np.Name = r.generateName("nodeplugin-csi-addons")
+	np.Namespace = r.driver.Namespace
```
Madhu-1: can we set the csiaddons ds as the ownerref to get it garbage collected?

PendingReleaseNotes.md
```diff
@@ -8,4 +8,5 @@
 - Fencing can now be enabled in the ceph-csi-drivers Helm chart either globally for all drivers or on a per-driver basis.
 - Added `enabled` field to log rotation configuration in the ceph-csi-drivers Helm chart. Users can now disable log rotation by setting `log.rotation.enabled: false` either globally in `operatorConfig.driverSpecDefaults` or per-driver in `drivers.<driver-type>`. This resolves Helm warnings when disabling file logging from parent charts.
 - Added support to set the priorityClass in the operator helm chart.
+- Added NetworkPolicies for the operator pod and CSI driver pods (controller-plugin, csi-addons node agent). Included in all generated manifests by default. Driver pod NPs are created by the operator for every reconciled driver. Node-plugin pods are exempt (`hostNetwork: true`).
```
Madhu-1: can you clear the above release notes , it was missed

internal/controller/driver_controller.go
```diff
@@ -1044,6 +1049,61 @@ func (r *driverReconcile) reconcileControllerPluginDeployment() error {
 	return err
 }
 
+func (r *driverReconcile) reconcileControllerPluginNetworkPolicy() error {
+	np := &networkingv1.NetworkPolicy{}
+	np.Name = r.generateName("ctrlplugin")
+	np.Namespace = r.driver.Namespace
+
+	log := r.log.WithValues("networkPolicy", np.Name)
```
Madhu-1: a condition check is missing to create this only when required not always

config/network-policy/operator-network-policy.yaml
```diff
@@ -0,0 +1,14 @@
+apiVersion: networking.k8s.io/v1
+kind: NetworkPolicy
+metadata:
+  name: ceph-csi-controller-manager
```
Madhu-1, line 4: this is not added to helm charts

docs/design/network-policy.md
Rakshith-R: NP creation for ctrl-plugin  is now gated with this.
host network pods cannot be restricted using NP.

config/network-policy/operator-network-policy.yaml
Rakshith-R: added now

internal/controller/driver_controller.go
```diff
@@ -1044,6 +1049,77 @@ func (r *driverReconcile) reconcileControllerPluginDeployment() error {
 	return err
 }
 
+func (r *driverReconcile) reconcileControllerPluginNetworkPolicy() error {
+	np := &networkingv1.NetworkPolicy{}
+	np.Name = r.generateName("ctrlplugin")
+	np.Namespace = r.driver.Namespace
+
+	pluginSpec := cmp.Or(r.driver.Spec.ControllerPlugin, &csiv1.ControllerPluginSpec{})
+	if ptr.Deref(pluginSpec.HostNetwork, false) {
```
Madhu-1, line 1058: we also need to delete when no csiaddons and no metadata features are enabled or they disabled later

internal/controller/driver_controller.go
```diff
@@ -1260,6 +1336,62 @@ func (r *driverReconcile) reconcileNodePluginDaemonSetForCsiAddons() error {
 	return err
 }
 
+func (r *driverReconcile) reconcileNodePluginCsiAddonsNetworkPolicy() error {
+	np := &networkingv1.NetworkPolicy{}
+	np.Name = r.generateName("nodeplugin-csi-addons")
+	np.Namespace = r.driver.Namespace
+
+	if !ptr.Deref(r.driver.Spec.DeployCsiAddons, false) || r.isNfsDriver() {
```
Madhu-1: can you move isNfs check above to skip not not call delete?

internal/controller/driver_controller.go
```diff
@@ -1260,6 +1336,62 @@ func (r *driverReconcile) reconcileNodePluginDaemonSetForCsiAddons() error {
...
+	log := r.log.WithValues("networkPolicy", np.Name)
+	log.Info("Reconciling csi-addons nodeplugin network policy")
+
+	proto := corev1.ProtocolTCP
+	opResult, err := ctrlutil.CreateOrUpdate(r.ctx, r.Client, np, func() error {
+		if err := ctrlutil.SetControllerReference(&r.driver, np, r.Scheme); err != nil {
+			return err
+		}
+		ds := &appsv1.DaemonSet{}
+		ds.Name = np.Name
```
Madhu-1: is np name and the controllerplugin ds names are same?

internal/controller/driver_controller.go
```diff
@@ -1260,6 +1336,62 @@ func (r *driverReconcile) reconcileNodePluginDaemonSetForCsiAddons() error {
 	return err
 }
 
+func (r *driverReconcile) reconcileNodePluginCsiAddonsNetworkPolicy() error {
+	np := &networkingv1.NetworkPolicy{}
+	np.Name = r.generateName("nodeplugin-csi-addons")
+	np.Namespace = r.driver.Namespace
+
+	if !ptr.Deref(r.driver.Spec.DeployCsiAddons, false) || r.isNfsDriver() {
+		if err := r.Delete(r.ctx, np); client.IgnoreNotFound(err) != nil {
```
Madhu-1, line 1347: we dont need to call delete as it gets garbage collected when the csiaddons ds is deleted?

internal/controller/driver_controller.go
Rakshith-R: Yes, `hostNetwork` is a pod-spec field — all containers in the Deployment share the same network namespace. When `pluginSpec.HostNetwork = true`, the csi-addons sidecar also runs on host network. The updated code now skips NP creation entirely when `hostNetwork: true` since NetworkPolicy enforcement is bypassed for host-network pods.

internal/controller/driver_controller.go
Rakshith-R: Yes, both use `r.generateName("nodeplugin-csi-addons")` — intentional so the ownerRef Get lookup works with the same name.

internal/controller/driver_controller.go
Rakshith-R: > Yes, `hostNetwork` is a pod-spec field — all containers in the Deployment share the same network namespace. When `pluginSpec.HostNetwork = true`, the csi-addons sidecar also runs on host network. The updated code now skips NP creation entirely when `hostNetwork: true` since NetworkPolicy enforcement is bypassed for host-network pods.

claude does not know :laughing:  

@black-dragon74 I think we missed adding the csi-addons daemonset podnetwork hack to ctrlplugin (host networking mode). 
We need to add additional csi-addons deployment in pod networking with special pod affinity(tricky part) when ctrlplugin pods are in host networking.

Can you please confirm this? 

Regardless, this enhancement should be taken outside the scope of this pr.

internal/controller/driver_controller.go
Rakshith-R: Addressed — ctrl-plugin NP is now deleted when `len(ingress) == 0` (no csi-addons and no snapshot-metadata TLS configured). Also deleted when `hostNetwork: true`.

internal/controller/driver_controller.go
Rakshith-R: Done — NFS check is now the first early return, before any NP object construction or delete calls.

internal/controller/driver_controller.go
Rakshith-R: Kept the explicit delete because Driver CR is also set as controller owner (needed for the `Owns(&NetworkPolicy{})` watch). With two owners, K8s GC won't delete the NP when only the DS is removed. The DS ownerRef is still set for the common case where the Driver CR itself is deleted.

Rakshith-R:
I'll double check ctrlplugin NP creation/deletion logic and then ask for review.

Rakshith-R:
## Flow: `reconcileNodePluginCsiAddonsNetworkPolicy`

1. NFS → **return nil** (no NP needed, never was)
2. `!DeployCsiAddons` → **delete NP**, return
3. **CreateOrUpdate NP** — Driver = controller owner, ingress from csi-addons controller-manager → 9071, open egress

> csiaddons nodeplugin Daemonset is not added as owner since the controller needs to own the NP to be able to watch it. Therefore, to keep it simple, NP is explicitely deleted. 

---

## Flow: `reconcileControllerPluginNetworkPolicy`

1. `hostNetwork: true` → **delete NP**, return
2. Build ingress slice:
   - `DeployCsiAddons` → add rule: csi-addons controller-manager → csiAddonsPort
   - RBD + `tls-key` volume → add rule: any pod → 50051
3. **CreateOrUpdate NP** — Driver = controller owner, ingress (may be empty = deny-all), open egress

> Creating a CSIAddons deployment to pair with ctrlplugin ( in host network mode) makes sense to avoid exposing node port but is out of scope for this pr. It can be taken as another enhancement in follow-up pr.

internal/controller/driver_controller.go
Rakshith-R: > we also need to delete when no csiaddons and no metadata features are enabled or they disabled later

If there's no ingress that needs to be opened, all ingress will be restricted and egress will be left opened. NP is still required.

internal/controller/driver_controller.go
```diff
@@ -1260,6 +1334,58 @@ func (r *driverReconcile) reconcileNodePluginDaemonSetForCsiAddons() error {
...
+	np := &networkingv1.NetworkPolicy{}
+	np.Name = r.generateName("nodeplugin-csi-addons")
+	np.Namespace = r.driver.Namespace
+
+	if !ptr.Deref(r.driver.Spec.DeployCsiAddons, false) {
+		if err := r.Delete(r.ctx, np); client.IgnoreNotFound(err) != nil {
+			return err
+		}
+		return nil
+	}
```
Madhu-1, line 1351: we dont need this check right, this get deleted when the csiaddons daemonset gets deleted

internal/controller/driver_controller.go
```diff
@@ -1044,6 +1049,75 @@ func (r *driverReconcile) reconcileControllerPluginDeployment() error {
...
 }
 
+func (r *driverReconcile) reconcileControllerPluginNetworkPolicy() error {
+	np := &networkingv1.NetworkPolicy{}
+	np.Name = r.generateName("ctrlplugin")
+	np.Namespace = r.driver.Namespace
+
+	pluginSpec := cmp.Or(r.driver.Spec.ControllerPlugin, &csiv1.ControllerPluginSpec{})
+	if ptr.Deref(pluginSpec.HostNetwork, false) {
+		if err := r.Delete(r.ctx, np); client.IgnoreNotFound(err) != nil {
```
Madhu-1, line 1059: i dont see the code to delete the NP when the csiaddons and metadata sidecars are disabled.

Madhu-1: okay this is kind of extra check ,doesnt hurt

Madhu-1: wrong comment

internal/controller/driver_controller.go
Madhu-1: okay this is kind of extra check ,doesnt hurt

Rakshith-R: >csiaddons nodeplugin Daemonset is not added as owner since the controller needs to own the NP to be able to watch it. Therefore, to keep it simple, NP is explicitely deleted.

Explained here
https://github.com/ceph/ceph-csi-operator/pull/544#issuecomment-498841805711

Driver CR needs to own NP to reconcile and watch it. Adding Daemonset in between makes NP have two refs and little complicated. Adding delete is simple.

internal/controller/driver_controller.go
Madhu-1: what is the use of creating a NP if we have not opened any port?

Madhu-1: Please note the default policies should be applied by admin at the NS level not be the operator.

internal/controller/driver_controller.go
Rakshith-R: > i dont see the code to delete the NP when the csiaddons and metadata sidecars are disabled.

https://github.com/ceph/ceph-csi-operator/pull/544#discussion_r3588753495
my(a very proactive claude's) comment here was wrong. 

NP is still required to restrict all Ingresss as just a extra security layer even if the pod itself does not listen to any ports (and ofc allow all egress in case there's namespace wide deny all NP).

Madhu-1: This should be created by cluster admin at each namespace as a default, the operator should not be creating a default NP

internal/controller/driver_controller.go
Rakshith-R: https://github.com/ceph/ceph-csi-operator/pull/544#discussion_r3592976837 
latest comment here ,
NP is like a allowlist -> we still need one for the unrestricted egress part. 
(restricting ingresss ourselves is just extra proactive layer )

internal/controller/driver_controller.go
Rakshith-R: > This should be created by cluster admin at each namespace as a default, the operator should not be creating a default NP

We need NP for unrestricted egress for the controllerplugin (for kubeapi access).
It is not default.

Madhu-1 (approved):

PR #558, opened 2026-07-24
feat: add explict namespace to resources

iamsauravsharma:

This PR adds an explicit `namespace: {{ .Release.Namespace }}` to the metadata of all `Role` and `RoleBinding` resources in the Helm charts that were previously rendered without one.

`helm template` never injects a namespace into rendered manifests. `--namespace` flag only sets `.Release.Namespace` when present inside templates.  (see helm/helm#10737).

Tools that consume those output therefore needs to inject namespace themselves. kustomize did this until v5.8.0. Now it no longer injects the kustomization namespace into Helm-generated resources (kubernetes-sigs/kustomize#6058).  As a result, when chart is used via kustomize >= 5.8.0, the Roles/RoleBindings land in the default namespace instead of the release namespace.

Setting the namespace explicitly in the templates makes the charts independent of any namespace-injection behavior, which is the fix recommended for chart authors in the kustomize issue.

Yes, for all working installations. `helm install`/`helm upgrade` already places resources without an explicit namespace
  into the release namespace, so the rendered result is unchanged. kustomize < 5.8.0 with a kustomization `namespace:` overrides `metadata.namespace` regardless, so the final output is unchanged.

## Related issues ##

Refer: https://github.com/kubernetes-sigs/kustomize/issues/6058
Refer: https://github.com/helm/helm/issues/10737

## Future concerns ##

* None.

**Checklist:**

* [X] **Commit Message Formatting**: Commit titles and messages follow guidelines in the [developer guide](https://github.com/ceph/ceph-csi-operator/blob/main/docs/development-guide.md#commit-messages).
* [X] Reviewed the developer guide on [Submitting a Pull Request](https://github.com/ceph/ceph-csi-operator/blob/main/docs/development-guide.md#development-workflow)

### DIFF
--- a/deploy/charts/ceph-csi-drivers/templates/cephfs-ctrlplugin-r-rbac.yaml
+++ b/deploy/charts/ceph-csi-drivers/templates/cephfs-ctrlplugin-r-rbac.yaml
@@ -6,6 +6,7 @@ apiVersion: rbac.authorization.k8s.io/v1
 kind: Role
 metadata:
   name: {{ $normalizedDriverName }}-ctrlplugin-r
+  namespace: {{ $root.Release.Namespace }}
 rules:
 - apiGroups:
   - coordination.k8s.io
--- a/deploy/charts/ceph-csi-drivers/templates/cephfs-ctrlplugin-rb-rbac.yaml
+++ b/deploy/charts/ceph-csi-drivers/templates/cephfs-ctrlplugin-rb-rbac.yaml
@@ -6,6 +6,7 @@ apiVersion: rbac.authorization.k8s.io/v1
 kind: RoleBinding
 metadata:
   name: {{ $normalizedDriverName }}-ctrlplugin-rb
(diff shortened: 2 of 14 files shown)


Madhu-1:
@iamsauravsharma Thanks for the PR  i see that this is covering only the driver chart what about adding namespace to the operator chart as well?

iamsauravsharma:
@Madhu-1 Actually I first added namespace for service account from operator chart. But CI was failing I was just tinkering and testing operator service account to fix those issue. I have re added Service Account namespace as well along with hack script fix

Note `hack/patch-csi-operator-helm-chart.sh` fix are done by AI so take consider since I am not familiar with the helmify/hack script setup

Issue Action run: https://github.com/ceph/ceph-csi-operator/actions/runs/30102884891/job/90183406865?pr=558

Madhu-1 (approved):

PR #575, opened 2026-08-13
Disable CSI-Addons for NVMe-oF

nixpanic:
NVMe-oF and NFS do not implement any of the CSI-Addons remote procedure
calls. The CSI-Addons sidecar was not added to NFS already, and should
not be added to NVMe-oF deployments either.

See-also: red-hat-storage/ocs-client-operator#717

### DIFF
--- a/internal/controller/driver_controller.go
+++ b/internal/controller/driver_controller.go
@@ -840,7 +840,7 @@ func (r *driverReconcile) reconcileControllerPluginDeployment() error {
 							})
 						}
 						// Addons Sidecar Container
-						if !r.isNfsDriver() && ptr.Deref(r.driver.Spec.DeployCsiAddons, false) {
+						if r.withCsiAddons() {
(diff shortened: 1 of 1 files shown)


black-dragon74 (approved):

internal/controller/driver_controller.go
```diff
@@ -840,7 +840,7 @@ func (r *driverReconcile) reconcileControllerPluginDeployment() error {
 							})
 						}
 						// Addons Sidecar Container
-						if !r.isNfsDriver() && ptr.Deref(r.driver.Spec.DeployCsiAddons, false) {
+						if r.withCsiAddons() {
```
Madhu-1, line 843: can we have same check on reconcileNodePluginDaemonSetForCsiAddons?

nixpanic: sure, done!

Madhu-1 (approved):

Issue #577, opened 2026-08-14
Helm: default driver log rotation not compatible with default privileged value.


sshambar:
# Describe the bug #

Deploying the drivers with their default values results in driver controllers failing to start due to SELinux labeling on the log-dir.

# Environment details #

- Image/version of ceph-csi-operator : 1.0.4
- Image/version of Ceph-CSI driver : 3.17.0
- Kubernetes cluster version : 1.36.3
- Ceph cluster version : 20.2.1

# Steps to reproduce #

1. Deploy a driver using the helm chart with defaults for log.rotation.enabled (true) and controllerPlugin.privileged (false)
2. On any host with SELinux in enforcing mode, note that the driver controller fails to start.

# Additional context #

The controller pods are deployed with securityContext.privileged: false, but also have --log_file args referencing volumeMounts from hostPath volumes from the host's /var/lib/cephcsi/<XXX> directory.

The SELinux labels for the /var/lib/cephcsi/<dir> are either system_u:object_r:container_file_t:s0 (or system_u:object_r:var_lib_t:s0 after a system relabel, after for example a host package upgrade).

The process is running with a MCS context, eg system_u:system_r:container_t:s0:c321,c556, which is unable to access the hostPath.

Kubernetes does not relabel hostPath directories to prevent breaking host processes (only PVC's, projected volumes etc are relabeled).

The operator should probably ignore the privileged: false setting if hostPath log directories are used, or support explicit securityContext settings for the controller which could be used with an privileged initContainer to correctly set the SELinux labels for the hostPath.

abhinav-phi:
Hi! I'd like to work on this.
Plan: I'll fix the Helm chart's default driver log rotation so it's compatible with the reported setup, adjust the default values, and document the setting.
Could you please assign this to me? Thanks!

/assign take

PR #592, opened 2026-08-30
operator: run controller plugin privileged when log rotation is enabled (fixes Helm default incompatibility with SELinux)

abhinav-phi:
## Summary

This PR addresses the controller plugin startup problem reported in #577 when hostPath log rotation is used on SELinux-enforcing hosts. The original implementation enables privilege whenever log rotation is enabled. Madhu raised a valid objection: that changes security settings on every platform and ignores the existing explicit configuration option.

**Status: awaiting agreement on a narrower design.** The proposed revision below is not implemented. The only recent commit change adds an AI assistance trailer; the source tree is unchanged. The open review thread has not been marked resolved.

## Problem and existing design

The reported deployment uses log rotation with a hostPath log directory while the controller plugin is configured as unprivileged. Access to that directory can be denied by SELinux policy. This does not mean every deployment with log rotation needs privileged containers, nor does it establish that privileged containers are the only possible solution.

The existing [log rotation design](https://github.com/ceph/ceph-csi-operator/blob/main/docs/design/logrotate.md) already documents `controllerPlugin.privileged` as an explicit setting for SELinux platforms that require privilege to write to the hostPath. That setting should remain available rather than being deprecated or silently overridden.

## What the current branch changes

The current implementation runs the log-writing controller containers privileged whenever log rotation is enabled, even when `controllerPlugin.privileged` is explicitly false. It also marks that field deprecated and ignored. The branch includes corresponding unit tests, Helm values documentation, release notes, and generated API and installation artifacts.

These are the disputed changes under review, not an accepted solution. The branch has not yet been revised to restore the original explicit privilege semantics.

## Proposed revision

The proposal in the [review-thread reply](https://github.com/ceph/ceph-csi-operator/pull/592#discussion_r4034342768) is to remove both the deprecation and the unconditional escalation, retain the existing explicit privilege setting, and narrow the contribution to clear Helm configuration and troubleshooting guidance for SELinux deployments using hostPath log rotation. Non-SELinux defaults would remain unchanged.

This would be configuration guidance rather than automatic SELinux detection or a claim that the default installation now works on every platform. In a mixed cluster, administrators would also need to align controller pod placement with the intended nodes, because a Deployment template does not select a different security context after each pod is scheduled.

Please confirm whether this documentation-focused direction matches the intended design before the implementation is revised.

## Validation and limitations

The branch contains unit tests for the original behavior; those tests do not validate the proposed revision. No tests were rerun for the recent commit-message-only amendment, and no SELinux-enabled deployment was tested during this follow-up. DCO passed after the amendment, but the PR workflows require maintainer approval, so this is not a claim that the complete CI suite is green.

The existing sign-off is preserved. The added `Assisted-by: ZCode AI assistant (review and revision planning)` trailer records the verified scope of AI assistance during this follow-up.

## Related issue

Related to #577. Resolution remains pending review of the design and validation of any subsequent revision.


abhinav-phi:
Friendly note for reviewers: the CI workflows on this PR are currently in the `action_required` state — they need a maintainer's approval to run since this is my first contribution to this repository. Could someone please approve the workflow runs? Thank you!

Madhu-1:
> Friendly note for reviewers: the CI workflows on this PR are currently in the `action_required` state — they need a maintainer's approval to run since this is my first contribution to this repository. Could someone please approve the workflow runs? Thank you!

@abhinav-phi if its written using AI please add assisted-by or co-authored-by to the commit

api/v1/driver_types.go
```diff
@@ -241,9 +241,13 @@ type ControllerPluginSpec struct {
 	//+kubebuilder:validation:Optional
 	Resources ControllerPluginResourcesSpec `json:"resources,omitempty"`
 
-	// To enable logrotation for csi pods,
-	// Some platforms require controller plugin to run privileged,
-	// For example, OpenShift with SELinux restrictions requires the pod to be privileged to write to hostPath.
+	// To enable logrotation for csi pods.
+	// Deprecated: when log rotation is enabled, the operator always runs the
```
Madhu-1, line 245: i dont think we need any deprecation notice , this is as per the design https://github.com/ceph/ceph-csi-operator/blob/c88282b004301a8514c98751ebf115c7661dce8a/docs/design/logrotate.md not all platforms will have selinux enabled, it need to be set when selinux is enabled.

abhinav-phi: Agreed, Madhu. The deprecation notice and the unconditional privilege escalation are both too broad, so I propose dropping both changes.

The narrower revision would keep the existing `controllerPlugin.privileged` setting as an explicit opt-in for SELinux deployments that use hostPath log rotation, and would document that setting in the Helm values and in troubleshooting guidance. Non-SELinux deployments would keep their current unprivileged defaults. One limitation worth noting for mixed clusters: this is a per-driver deployment setting, so on SELinux hosts the controller pods would also need node placement rules, because the security context cannot be chosen per node after scheduling.

This proposal was prepared with AI assistance, and no code has been revised yet. Does this documentation-focused approach match your intended design? I will revise the implementation once we agree on the direction.

abhinav-phi:
Madhu, I have replied in the open review thread with a narrower proposal that preserves the existing `controllerPlugin.privileged` setting and limits the guidance to SELinux deployments that use hostPath log rotation. Please review that direction before I revise the implementation.

On the AI assistance trailer: I have amended the commit to add `Assisted-by: ZCode AI assistant (review and revision planning)` alongside the existing `Signed-off-by`. Only the commit message changed; the source tree is identical, so the tests have not been rerun. The pushed code still reflects the original approach, and I am deliberately leaving the review thread open until we agree on the design.

abhinav-phi:
@Madhu-1, before I change the implementation, could you explicitly confirm whether you agree with this narrower scope: remove the blanket privilege escalation and field deprecation, preserve the existing explicit `controllerPlugin.privileged` setting, and limit this PR to Helm configuration and troubleshooting guidance for SELinux deployments using hostPath log rotation? This would leave non-SELinux defaults unchanged and would not introduce automatic SELinux detection. If you expect a code-level behavior change instead, please clarify the intended privilege semantics. I will wait for your direction before revising the branch.

Madhu-1:
> @Madhu-1, before I change the implementation, could you explicitly confirm whether you agree with this narrower scope: remove the blanket privilege escalation and field deprecation, preserve the existing explicit `controllerPlugin.privileged` setting, and limit this PR to Helm configuration and troubleshooting guidance for SELinux deployments using hostPath log rotation? This would leave non-SELinux defaults unchanged and would not introduce automatic SELinux detection. If you expect a code-level behavior change instead, please clarify the intended privilege semantics. I will wait for your direction before revising the branch.

i would just go with documentation for this one.

abhinav-phi:
Confirming docs-only: revert escalation+deprecation, keep explicit setting + Helm/docs guidance - starting revision.

abhinav-phi:
@Madhu-1 docs-only revision pushed, please re-review.

HEAD: 34a7abb -> a3d35025 (normal push, no force).

Reverted behavior change per your direction (5725566891):
- internal/controller/driver_controller.go: back to explicit pluginSpec.Privileged logic (pluginSpec.Privileged != nil && logRotationEnabled, Privileged: pluginSpec.Privileged), removed SELinux auto-escalation comment.
- api/v1/driver_types.go: restored original 3-line privileged comment, no deprecation notice (per design docs/design/logrotate.md).
- Deleted internal/controller/driver_controller_securitycontext_test.go (covered rejected behavior).
- Restored generated: config/crd/bases/*, deploy/all-in-one/*, deploy/multifile/crd.yaml, deploy/charts/ceph-csi-operator/templates/*, vendor/.../driver_types.go, PendingReleaseNotes.md escalation paragraph removed.

Docs-only (no auto-detection logic, API unchanged, privileged:false default kept):
- deploy/charts/ceph-csi-drivers/values.yaml: document set controllerPlugin.privileged:true with log.rotation.enabled/logHostPath for SELinux+hostPath only, cite docs/design/logrotate.md; explicitly states operator does not auto-escalate.
- docs/helm-charts/drivers-chart.md (+ .gotmpl.md source): synced table descriptions, added Troubleshooting log rotation on SELinux-enforcing hosts with mixed-cluster node-placement caveat (controllerPlugin.affinity/tolerations), no auto-privileged claim.

Verify: git diff --check clean, go build ./internal/controller/... passes, no Privileged:ptr.To(true) auto-escalation remains (only pre-existing node/sidecar privileged).

Thanks!

docs/helm-charts/drivers-chart.gotmpl.md
```diff
@@ -69,6 +69,23 @@ The following table lists the configurable parameters of the ceph-csi-drivers ch
 
 {{ template "chart.valuesTable" . }}
 
+### Troubleshooting log rotation on SELinux-enforcing hosts
+
+If log rotation (`log.rotation.enabled`) uses a hostPath volume (`log.rotation.logHostPath`)
+on hosts with SELinux in enforcing mode (for example, OpenShift), the controller plugin
+containers need `controllerPlugin.privileged:true` to write the rotated log files, as
+Kubernetes does not relabel hostPath volumes for SELinux. See `docs/design/logrotate.md`.
```
Madhu-1: provide a complete link to it

deploy/charts/ceph-csi-drivers/values.yaml
```diff
@@ -93,6 +93,9 @@ operatorConfig:
       verbosity: 0
       rotation:
         # -- Enable log rotation (default: true)
+        # On SELinux-enforcing hosts, combine with controllerPlugin.privileged:true
+        # when using hostPath log storage; the operator does not auto-escalate.
+        # See docs/design/logrotate.md.
```
Madhu-1: remove the link to the document as it will not be helpful from here

abhinav-phi:
@Madhu-1 Done in f6072946 — the troubleshooting section in drivers-chart.gotmpl.md now uses the complete design-doc URL, and all the See-docs lines are removed from values.yaml (operational guidance kept). drivers-chart.md is hand-synced to match since I don't have helm-docs locally — CI will confirm it matches a regen. PTAL, thanks!

Madhu-1:
@abhinav-phi can you please squash the commits to 1, changes LGTM

abhinav-phi:
@Madhu-1 Done — squashed to a single commit (45d9aa7c) and rebased onto current upstream/main. Net diff is docs-only as approved: values.yaml + drivers-chart.gotmpl.md + drivers-chart.md (84+/15-). DCO intact, git diff --check clean. PTAL for approve/merge when convenient. Thanks!

Issue #578, opened 2026-08-14
Create PodMonitor resource to automate Prometheus monitoring


nixpanic:
# Describe the feature you'd like to have #

Each Ceph-CSI Pod should have a PodMonitor, so that Prometheus knows how to scrape the CSI metrics.

Some considerations:
- Pods with hostnetworking (by default the node-plugins) can easily cause conflicts when listening on a local port
- ...

## Who is the end user and what is the use case where this feature will be valuable? ##
  
Through the Prometheus Operator it is possible to get statistics and setup alerting. This helps administrators identifying issues.

# How will we know we have a good solution? (acceptance criteria) #

Ceph-CSI metrics (from the standard Kubernetes CSI sidecars) become available in Prometheus.

# Additional context #

Suggested in ceph/ceph-csi#6287. Rook supports ServiceMonitor for the same purpose.
More details in https://prometheus-operator.dev/docs/developer/getting-started/ .

abhinav-phi:
Hi! I'd like to take this issue.
Plan: I'll add a PodMonitor resource to automate Prometheus scraping for the operator/CSI pods, make it optional via CR/flag, and add example manifests + docs.
Could you please assign this to me? Thanks!

/assign take

PR #593, opened 2026-08-30
operator: create PodMonitor resources to automate Prometheus monitoring of CSI metrics

abhinav-phi:
## Summary

This PR addresses #578 by adding operator-managed Prometheus `PodMonitor` resources for CSI workloads. The current implementation discovers the metrics endpoint of the Ceph-CSI `liveness-prometheus` sidecar. Madhu pointed out that this sidecar is deprecated and is expected to be removed, so it should not be the foundation of new monitoring support.

**Status: awaiting agreement on a controller-only alternative.** The alternative described below is a proposal, not the implementation currently on this branch. The recent commit amendment adds an AI assistance trailer without changing the source tree.

## What the current branch implements

The branch adds an optional `spec.podMonitor` configuration with enablement, labels, annotations, and scrape interval settings. These settings can also be supplied through `OperatorConfig` driver defaults. When enabled with liveness configured, reconciliation creates or updates a Driver-owned PodMonitor selecting that driver's controller and node plugin pods and scraping their named `metrics` port at `/metrics`.

The current implementation declares that port on the liveness sidecars and removes the PodMonitor when monitoring is disabled or liveness configuration is removed. It includes the monitoring API dependency, RBAC, generated CRDs and installation manifests, Helm configuration, documentation, and reconciliation tests. The external PodMonitor CRD is included for envtest rather than installation into user clusters.

Monitoring is opt-in. The implementation is intended to tolerate an absent PodMonitor CRD when monitoring is disabled; enabling the feature requires the Prometheus Operator CRD. Its dependence on `spec.liveness` is the architectural issue that remains unresolved.

## Proposed replacement for the liveness dependency

The [discussion reply](https://github.com/ceph/ceph-csi-operator/pull/593#issuecomment-5710856820) proposes starting with operation metrics from the controller's upstream Kubernetes CSI sidecars: external-provisioner, external-attacher, external-resizer, and csi-snapshotter when present. These support `--http-endpoint`, with `/metrics` as the default metrics path. The operator already permits per-container arguments through `spec.controllerPlugin.containerExtraArgs`, but arguments alone are not sufficient to complete monitoring integration.

The proposed revision would provide distinct named metrics ports for the enabled sidecars, matching NetworkPolicy ingress, and a PodMonitor that is independent of liveness configuration. Monitoring would remain opt-in, and ports would be configurable to account for shared pod networking and controller deployments using host networking. The precise configuration interface and supported image versions would need to be checked as part of implementation.

These endpoints expose CSI operation metrics; they are not an equivalent replacement for the old `csi_liveness` gauge. Node plugin monitoring would be excluded from this revision because the node driver registrar's health endpoint is not an equivalent operation-metrics source. The operator's own existing metrics monitoring is separate from this proposal.

## Corrections to the original description

[ceph-csi PR #6287](https://github.com/ceph/ceph-csi/pull/6287) is still open, so it should be treated as related discussion rather than established upstream Helm support. Also, host networking does not inherently rule out a ServiceMonitor; the choice here is direct pod discovery, not a claim that Service-based discovery is impossible.

## Validation and review request

The original submission reported passing lint, chart, and reconciliation checks, with an envtest teardown failure on Windows. Those results have not been independently rerun during this follow-up and do not validate the proposed controller-only design. No tests were rerun for the commit-message-only amendment. DCO passed, but that alone does not establish that the complete CI suite has passed.

The sign-off is preserved, and the added `Assisted-by: ZCode AI assistant (review and revision planning)` trailer records the verified assistance scope. Would this controller-only direction address the architectural concern, or is a different upstream metrics path preferred?

## Related issue

Related to #578. The current liveness-based implementation remains under review and has not been replaced by the proposed design.


Madhu-1:
liveness sidecar is deprecated and it will be removed

abhinav-phi:
Agreed, Madhu. This should not depend on the deprecated liveness sidecar, so here is a narrower proposal.

Instead of scraping the `liveness-prometheus` container, the PodMonitor would scrape the metrics endpoints of the upstream Kubernetes CSI sidecars on the controller plugin. The external-provisioner, external-attacher, external-resizer and csi-snapshotter sidecars all support `--http-endpoint` (with `/metrics` served by default), and the operator already supports per-container arguments through `spec.controllerPlugin.containerExtraArgs`. The revision would declare separate named ports for the enabled sidecars, add matching NetworkPolicy ingress, and keep the feature opt-in with configurable ports, which also matters when controller pods run with host networking.

To be clear about what this collects: these are CSI operation metrics from the controller sidecars, not a replacement for the old `csi_liveness` gauge. I would leave node plugin monitoring out of this revision, because the node driver registrar's HTTP endpoint serves only `/healthz`, not equivalent metrics.

One correction to my earlier description: ceph-csi PR #6287 (PodMonitor support in the Helm charts) is still open, so I should not have described it as established upstream support.

Would this controller-only scope address the architectural concern, or would you prefer a different upstream metrics path? This proposal was prepared with AI assistance.

Madhu-1:
> Agreed, Madhu. This should not depend on the deprecated liveness sidecar, so here is a narrower proposal.
> 
> Instead of scraping the `liveness-prometheus` container, the PodMonitor would scrape the metrics endpoints of the upstream Kubernetes CSI sidecars on the controller plugin. The external-provisioner, external-attacher, external-resizer and csi-snapshotter sidecars all support `--http-endpoint` (with `/metrics` served by default), and the operator already supports per-container arguments through `spec.controllerPlugin.containerExtraArgs`. The revision would declare separate named ports for the enabled sidecars, add matching NetworkPolicy ingress, and keep the feature opt-in with configurable ports, which also matters when controller pods run with host networking.
> 
> To be clear about what this collects: these are CSI operation metrics from the controller sidecars, not a replacement for the old `csi_liveness` gauge. I would leave node plugin monitoring out of this revision, because the node driver registrar's HTTP endpoint serves only `/healthz`, not equivalent metrics.
> 
> One correction to my earlier description: ceph-csi PR #6287 (PodMonitor support in the Helm charts) is still open, so I should not have described it as established upstream support.
> 
> Would this controller-only scope address the architectural concern, or would you prefer a different upstream metrics path? This proposal was prepared with AI assistance.

It would be fine to add this yamls to a separate directory and document this feature and how it can be used rather than adding this to the API, we already have helm support to create any arbitrary yamls as required that can be used for this one.

abhinav-phi:
Confirming: rewrite to separate-dir example yamls + usage docs, dropping API/RBAC/CRD surface, using Helm arbitrary-yaml - starting rewrite, will push.

abhinav-phi:
@Madhu-1 rewritten per your direction, please re-review.

New HEAD: b9815e6 (was 8bf1a2c).

Dropped: Driver PodMonitor API (api/v1, deepcopy), RBAC (monitoring.coreos.com podmonitors), CRD bases and generated manifests (all-in-one, multifile, charts), controller reconcile and tests, csi utils helpers, go.mod/go.sum dep and all vendor/prometheus-operator, drivers-chart docs table entries.

Added:
- deploy/examples/podmonitor/rbd-controller-podmonitor.yaml and cephfs-controller-podmonitor.yaml - controller-plugin sidecar metrics (provisioner/attacher/resizer/snapshotter via --http-endpoint, targetPort, /metrics), no liveness sidecar dependency, distinct ports per driver with hostNetwork conflict note.
- docs/design/podmonitor.md rewritten as usage-only guide (prereqs: Prometheus Operator, containerExtraArgs example, kubectl apply, Helm extraDeploy arbitrary-manifest example).
- PendingReleaseNotes.md docs-only entry; mkdocs nav kept for usage doc.

Verified: no PodMonitorSpec or prometheus-operator refs remain in api/internal/config/vendor/cmd, go build ./internal/controller/... passes, git diff --check clean. make generate/manifests skipped (heavy); CI approval still needed.

abhinav-phi:
Fixed DCO: added missing Signed-off-by to the rewrite commit (da9268ed) - DCO now passes. This is the docs/examples-only head you asked for (no operator API/RBAC/CRD, no liveness dependency, controller sidecars only, hostNetwork note). Please re-review when you get a chance. Thanks!

Madhu-1 (commented):
there a no need to have design document and multiple example for the different drivers, as its going to remain same for all drivers, just add one example and add a document to create multiple resources as required.

add a pre-req to the document, you can follow similar to https://github.com/ceph/ceph-csi-operator/blob/main/docs/features/rbd-snapshot-metadata.md

abhinav-phi:
@Madhu-1 Friendly ping - docs/examples-only rewrite per your direction (DCO fixed), HEAD b9815e6/da9268ed. Please re-review when convenient. Thanks!

PR #607, opened 2026-09-10
ci: use docker as container runtime

Madhu-1:
in ubuntu 24 the containerd is having the deprecated API because of the minikube cluster creation is failing, as we are installing the docker-cri, defaulting the minikube to pick the docker as the container runtime


Madhu-1:

